PR #998, opened 2023-11-03
Allow franken to set names

scp410:


jamesrudolph (commented):
First glance this looks alright, but, without looking (only on mobile rn) I fear there are a lot of Mapper.isValidFaction() checks that will fail.
I'll pull this down tomorrow and take a closer look if nobody else does before me

jamesrudolph:
Yeah, a few places this would fail:
https://github.com/search?q=repo%3AAsyncTI4/TI4_map_generator_bot%20isFaction&type=code

jamesrudolph:
Done with #1585

PR #1015, opened 2023-11-07
AC/PN/Tech/Leader Temp combat mods now working

jrkd:
Modifiers can be 1 round, 1 combat or 1 tactical action.

Also added a mod with condition to check whether your the owner of a tekklar PN that's currently in play, which cancels out the +1 that norr gets

These are saved against the player in the save file, in case of server going down after applying


jrkd:
Currently only handles combatround based rolls, which means Bunker doesnt work yet. 

Also probably need to consider the temp-extra rolls to allow for Experimental space station etc

jamesrudolph (approved):

PR #1022, opened 2023-11-08
Little Omega and ACD2 updates

wholton:


src/main/resources/data/action_cards/action_cards.json
```diff
@@ -1587,7 +1587,7 @@
     "alias": "boarding_torpedoes",
     "name": "Boarding Torpedoes",
     "phase": "Action",
-    "window": "After you destroy an opponent's non-fighter ship",
+    "window": "After a hit is assigned to your opponent's non-fighter ship",
```
jamesrudolph, line 1590: This no longer fits the theme of "Boarding Torpedoes" - how does the boarding crew force a Dread to sustain damange, and also spawn a brand new Dread?

wholton: I *believe* Sustain cancels before hits are assigned.

PR #1089, opened 2023-11-20
Update base.json

newnew1450:
Remove Valefar techs from Nekro starting Tech. They are already included in their faction techs, serves no specific purpose, and messes up franken draft.


jamesrudolph:
The might be listed in starting techs to control behavior reliant on having those techs, but as of right now, I don't see anywhere in the code where we are searching for those techs. Everything seems based on **technological singularity** and **galactic threat** and we are not yet validating whether a faction tech can be taken or not.

PR #1171, opened 2023-12-07
Fix/reenable franken edit for bothelpers

tedw4rd:


Fingon00:
I merged this under a different pull request

PR #1184, opened 2023-12-10
Issue 1023 fix winnu flagship

BlizzardOf:
The Winnu flagship's intended behavior is to roll a die for every non-fighter ship that it is fighting. I handled this by giving it a base die value of 0, then adding a new modifier that adds an extra die for every enemy non-fighter. I needed to obtain the opponent's ships at the CombatRoll level and pass it down to the roll modifier calculator in order to determine how many non-fighters are in the combat.


jamesrudolph (approved):
Ran a quick test on my local - seems to work!

PR #1223, opened 2023-12-15
Basic JSON serialization tests for `UnitKey`

aaronbruckner:
# Description

More serialization tests, this time for `UnitKey`.

`UnitKey` is causing us issues as it's being used as a key in Java maps which messes up serialization. JSON only allows map keys to be strings, causing issues we we attempt to deserialize back to a Java map.

This ensures we can at least serialize UnitKey which was taking on tons of extra fields that I had to drop.

Next step will be to modify the serialization process to get this working when `UnitKey` is used as a map key.

# Testing

```
docker build -t tibot .
```

### DIFF
--- a/src/main/java/ti4/helpers/Units.java
+++ b/src/main/java/ti4/helpers/Units.java
@@ -1,6 +1,10 @@
 package ti4.helpers;
 
 import java.util.concurrent.ThreadLocalRandom;
+
+import com.fasterxml.jackson.annotation.JsonIgnore;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
 import lombok.Data;
 import lombok.Getter;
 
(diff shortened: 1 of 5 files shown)


jamesrudolph (approved):
neat!

So essentially every field we expect to serialize should be in the `Set<String> knownJsonAttributes` and if anything gets added afterwards, without explicitly adding it to the test, will fail?

Even if `Game` is monolithic, I think this will help prevent most json serialization issues I've seen crop up over time

aaronbruckner:
@jamesrudolph yep! `knownJsonAttributes` is a hard allow list. That way we don't accidentally add fields we never intended to save and whenever we add a new field, it forces the IC to actually update the test.

PR #1224, opened 2023-12-15
Add bothelper send_games_to_s3 command

adam133:


jamesrudolph:
Sorry @adam133 - I wasn't very clear.
I `Game.getGameWinner().isEmpty == true` means there is not a winner.
`isPresent()` is what I should have recommended.

And I *would* couple that with hasEnded(), on the off chance there are games that people haven't ended yet for whatever reason

adam133:
Doh. If I had thought about that for more than a half second...

> Sorry @adam133 - I wasn't very clear. I `Game.getGameWinner().isEmpty == true` means there is not a winner. `isPresent()` is what I should have recommended.
> 
> And I _would_ couple that with hasEnded(), on the off chance there are games that people haven't ended yet for whatever reason

Doh. Should have thought about that for more than half a second. Should be fixed?

PR #1228, opened 2023-12-16
Space JSON Serialization with Map-key special logic

aaronbruckner:
# Description

Because `UnitKey` objects are used as keys within Java maps, we need special serialization logic for them as JSON map keys can only be strings. So we must make the "key" a JSON string which we then unwrap when deserializing.

Before this, Jackson was using `UnitKey.toString()` to convert the object key into a string which is not at all what we want.

I tried getting annotations working to add the key serializers/deserializers but I could never get the deserializer to trigger. Only way I got it working was to attach them to the `ObjectMapper`.

At some point we will need to do a pass and update all `ObjectMapper`'s to use the factory I created. Otherwise they will fail to convert objects.

# Testing

```
docker build -t tibot .
```

### DIFF
--- a/src/main/java/ti4/map/UnitHolder.java
+++ b/src/main/java/ti4/map/UnitHolder.java
@@ -2,10 +2,10 @@ package ti4.map;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.fasterxml.jackson.annotation.JsonSubTypes;
 import com.fasterxml.jackson.annotation.JsonTypeInfo;
+
 import java.awt.Point;
 import java.util.HashMap;
 import java.util.HashSet;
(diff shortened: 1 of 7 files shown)


src/main/java/ti4/map/UnitHolder.java
```diff
@@ -17,11 +17,10 @@
 import ti4.helpers.Units.UnitKey;
 import ti4.helpers.Units.UnitType;
 
-@JsonIgnoreProperties(ignoreUnknown = true)
-@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "type")
+@JsonTypeInfo(use = JsonTypeInfo.Id.NAME, property = "javaClassType")
```
aaronbruckner, line 20: Made this more explicit so we don't collide with similar type-like properties.

src/main/java/ti4/map/UnitHolder.java
```diff
@@ -176,6 +179,7 @@ public boolean hasUnits() {
         return !getUnits().isEmpty();
     }
 
+    @JsonProperty("unitsDamage")
     public HashMap<UnitKey, Integer> getUnitDamage() {
         return unitsDamage;
     }
```
aaronbruckner, line 185: Made this plural as I think that is more accurate and we will probably want to refactor the function name eventually.

PR #1313, opened 2024-01-06
Add flavor text for PoK, Codex II relics

### LINKED ISSUE #1040: Add Relic Flavour Text
Relics are missing their Flavour Text:
https://github.com/TerTerro/TI4_map_generator_bot/blob/858c2f950bee8313b04ab909454edce5b7196d5a/src/main/resources/data/relics/pok.json
https://github.com/TerTerro/TI4_map_generator_bot/blob/5c2f2e87517221d49dbc13319e2f52e0b207afbe/src/main/resources/data/relics/codexii.json
https://github.com/TerTerro/TI4_map_generator_bot/blob/5c2f2e87517221d49dbc13319e2f52e0b207afbe/src/main/resources/data/relics/ds.json

Copy how Absol's has it:
https://github.com/TerTerro/TI4_map_generator_bot/blob/a40e3f5c6680032610f2c342fe3831eac81333c1/src/main/resources/data/relics/absol.json

ninevra:
Fixes #1040.

### DIFF
--- a/src/main/resources/data/relics/codexii.json
+++ b/src/main/resources/data/relics/codexii.json
@@ -3,7 +3,8 @@
         "alias": "dynamiscore",
         "name": "Dynamis Core",
         "text": "While this card is in your play area, your commodity value is increased by 2. Action: Purge this card to gain trade goods equal to your printed commodity value+2.",
-        "source": "codex2"
+        "source": "codex2",
+        "flavourText": "Dart and Tai stood, mesmerized by the swirling mass. Neither had any idea what it was, but both were fairly sure they were about to retire."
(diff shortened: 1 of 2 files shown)


ninevra:
This does not properly render the emphasis in The Prophet's Tears ("*...and you think* I'm *being irrational?*").

The rendering code italicizes flavor text by wrapping it in `*`, and nesting emphasis in markdown is nontrivial (to be polite), so rendering that properly would require either putting a workaround in the data (i.e. writing `think* I'm *being` instead of the semantically-correct `think *I'm* being`) or some complicated logic.

PR #1345, opened 2024-01-16
Handle empty whispers

### LINKED ISSUE #1270: Index out of bounds
https://github.com/AsyncTI4/TI4_map_generator_bot/blob/master/src%2Fmain%2Fjava%2Fti4%2FMessageListener.java#L520

Use substringBetween here and then null check the string before using it 

Error here https://discord.com/channels/943410040369479690/1012679443615064104/1188036573380476938

ninevra:
Closes #1270.

I had to do at least a little refactoring in order to read the code, but i tried to keep it to a minimum.

Changes the substring logic to produce empty strings instead of StringIndexOutOfBoundsExceptions. This lets empty whispers be sent and received without error.

Adds a check to catch empty whispers and log them instead of sending them.

~~(I think not sending them is better, because otherwise accidentally omitting the space between the address and the message could trigger one of the more confusing bugs.)~~ (Nevermind, only makes a difference for single-word messages.)

### DIFF
--- a/src/main/java/ti4/MessageListener.java
+++ b/src/main/java/ti4/MessageListener.java
@@ -505,12 +505,16 @@ public class MessageListener extends ListenerAdapter {
         }
 
         if (messageToColor || messageToMyself || messageToFutureColor || messageToJazz) {
+            final String messageContent = StringUtils.substringAfter(messageText, " ");
+
             String gameName = event.getChannel().getName();
             gameName = gameName.replace("Cards Info-", "");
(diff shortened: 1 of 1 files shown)


ninevra:
Opened an issue #1346 collecting bugs i found while (manually, unfortunately) testing this. i could address those in a further refactor if desired.

jamesrudolph:
I had an idea for the empty whisper last night.

What if an empty whisper just reports back to the user:
"You tried to send an empty whisper. This whisper would have gone to XYZ player."

PR #1459, opened 2024-02-22
block unloadable emojis from being used as franken icons

tedw4rd:


tedw4rd:
@Fingon00 heads up, this should prevent the issue we were talking about last night.

PR #1504, opened 2024-03-08
Add second colour for planetary shield grids

bradleysigma:
Adds an orange version of the planetary shield bubble grid. Planets where the blue grid didn't contrast enough will now use the orange grid. Alas, for some planets, neither grid contrasts, but adding additional grid colours may cause more confusion that it alleviates. Also, some bubble grids were misaligned on their planets; this has been adjusted.
pbd4_2024 03 08 - 06 46 40
It's been a few years since I coded in Java, and this is my first real contribution to this project, so please let me know if I've erred in style conventions and the like.


bradleysigma:
I'll also note that the orange bubble grid has only been applied to the non-homebrew planets.

PR #1519, opened 2024-03-13
Eronous is bored

Eronous:
Its back, with fewer spelling mistakes


src/main/resources/systems/er11.json
```diff
@@ -5,7 +5,7 @@
         "asteroidbeta",
         "asteroidbwh"
     ],
-    "imagePath": "ER11_betaasteroid.png",
+    "imagePath": "er)11_async.png",
```
jamesrudolph, line 8: this looks wrong

src/main/resources/planets/rialarhon.json
jamesrudolph, line 1: changing the ID will break all existing "rialarhon" but maybe that's fine?

src/main/resources/systems/er68.json
```diff
@@ -1,11 +1,11 @@
 {
     "id": "er68",
-    "name": "Nebula2e",
+    "name": "Supernova2e",
```
jamesrudolph, line 3: what's the "2e" all about?

Eronous: 2e just means second [tile] eronous

src/main/resources/planets/rialarhon.json
Eronous: I dont know what it will do

src/main/resources/systems/er11.json
Eronous: er_11_async.png

PR #1579, opened 2024-03-26
resized icons

bradleysigma:
All icons in `\src\main\resources\emojis\factions\` have been squared, and resized to 90x90. Also, the images are now properly `.png` files, instead of `.webp` files labelled with a `.png` extension.


bradleysigma:
Also, `Cabal.png` was in the `base` folder, instead of the `pok` folder; I have left it there for now.

jamesrudolph:
I'm hesitant to merge these because they are for emojis and not the 95x95 faction icons used by the bot.
In an ideal world the emojis would be square, 128x128 or higher, and be a maximum 256kb - I worry with the 90x90 we would lose quality in places we grab the image from the emoji URL

jamesrudolph:
there's been a big switchover to higher scale faction images to support start of turn banners - going to close this

PR #1633, opened 2024-04-04
Promises Promises & Flagshipping

jamesrudolph:


src/main/resources/data/units/flagshipping.json
```diff
@@ -0,0 +1,1029 @@
+[
+    {
+        "id": "sigma_sol_flagship_1",
+        "baseType": "flagship",
+        "name": "Genesis I",
+		"asyncId": "fs1",
```
jamesrudolph, line 6: asyncID is used to know which unit image to place, typically this would be "fs"
We could potentially look at two different images to differentiate the upgraded flagship

src/main/java/ti4/map/Game.java
```diff
@@ -161,6 +161,12 @@ public class Game {
     private boolean miltyModMode;
     @Getter
     @Setter
+    private boolean promisesPromises;
+    @Getter
+    @Setter
+    private boolean flagshipping;
+    @Getter
+    @Setter
```
jamesrudolph, line 169: GameSaveLoadManager.java will need to explicitly save these values, akin to other booleans

jamesrudolph:
@bradleysigma minor changes to review

bradleysigma:
I've attempted to make the requested changes, though given that this is the first time I've programmed in Java in about ten years, they should probably be thoroughly reviewed.

bradleysigma:
@jamesrudolph, could you please review this again?

PR #1680, opened 2024-04-17
Change `planetTypes` to `planetType` for kytos, prymis

ninevra:
Without a `planetType`, `PlanetModel#createRepresentationEmbed()` throws when trying to switch over the null returned from `getPlanetType()`. This breaks `/search planets` when kytos or prymis is in the search results.

Could add a `@NonNull` annotation to the `planetType` field to catch this kind of data error, or could change `createRepresentationEmbed()` to handle nulls if this kind of data is intentional. Am happy to implement either.


jamesrudolph:
I think @Jazzxhands was playing with PlanetType recently - if NO planet should have nulls, we should add the check in PlanetModel.isValid(), otherwise getPlanetType should return an Optional<PlanetType> to follow suit how other nulls are handled

PR #1701, opened 2024-04-20
Add Peeked Markers to Map

BurningSkyMK1:
This adds the ability to track what player(s) have peeked at an unrevealed public objective, either stage 1 or 2, and draw markers onto the map to display this information.

The goal of this feature is to better convey to the table which public objectives player(s) have used the Espionage SC on.


bradleysigma:
I would suggest changing each instance of "peak" (mountain top) to "peek" (glance).

BurningSkyMK1:
Done @bradleysigma. I only changed the code I directly worked with from 'peaked' to 'peeked'. I.e., I didn't update `publicObjectives1Peakable` or `publicObjectives2Peakable`, as that touches code outside the scope of my changes.

src/main/java/ti4/AsyncTI4DiscordBot.java
```diff
@@ -375,6 +377,7 @@ private static void initializeWhitelistedRoles() {
         bothelperRoles.add(jda.getRoleById("1063464689218105354")); // FoW Server Game Admin
         bothelperRoles.add(jda.getRoleById("1131925041219653714")); //Jonjo's Server
         bothelperRoles.add(jda.getRoleById("1215450829096624129")); //Sigma's Server
+        bothelperRoles.add(jda.getRoleById("1225597399385374781")); //Sigma's Server
```
jamesrudolph, line 380: Forgot to update comment

BurningSkyMK1: Oops, good catch @jamesrudolph.

BurningSkyMK1: Fixed.

PR #1702, opened 2024-04-21
Send Assasination msg to targets private thread in fow mode

Solaxi:
Resolving Assassinate Representative AC in FOW-mode shows the message only to the resolver:
image
Changed so that the 2nd line is sent to the targets private thread instead.


Fingon00:
Wrapped this PR into a different one, nice catch

PR #1722, opened 2024-04-28
Holy tispoon attempts

HolyTispoon:


src/main/java/ti4/buttons/ButtonListener.java
```diff
@@ -4621,6 +4621,7 @@ public void resolveButtonInteractionEvent(ButtonInteractionEvent event) {
                 }
                 case "mitosisMech" -> ButtonHelperAbilities.resolveMitosisMech(buttonID, event, activeGame, player,
                     ident, finsFactionCheckerPrefix);
+                case "searchMyGames" -> SearchMyGames.SearchMyGames (player, event);
```
HolyTispoon, line 4624: I know this is wrong but couldn't find anything in the SearchMyGames command that looked like the simpleShowGame or sendCardsInfo of the other button listeners.

Fingon00: Normally I just make those out of the bones of the class

PR #1736, opened 2024-05-02
Don't check for private channels for non-real players in FoW

Solaxi:
Every time GM presses buttons in FoW, this gets shown:
image
Set the gm-room channel as the private channel for GM.

If GM is also a player, then this gets replaced when player private channels are setup.

### DIFF
--- a/src/main/java/ti4/buttons/ButtonListener.java
+++ b/src/main/java/ti4/buttons/ButtonListener.java
@@ -193,7 +193,7 @@ public class ButtonListener extends ListenerAdapter {
             activeGame.increaseButtonPressCount();
 
             if (activeGame.isFoWMode()) {
-                if (player != null && player.getPrivateChannel() == null) {
+                if (player != null && player.isRealPlayer() && player.getPrivateChannel() == null) {
                     MessageHelper.sendMessageToChannel(event.getChannel(),
                         "Private channels are not set up for this game. Messages will be suppressed.");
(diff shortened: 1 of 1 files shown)


Fingon00 (commented):
Sounds like the message is wrong to be sent in the first place, and should be curtailed/modified/possibly entirely removed. Setting up a private channel for a non-real player (the GM) seems like it could lead to confusion/unintended consequences down the line

Fingon00:
Also private channel has a pretty explicit definition, it ends in -private, and the GM room is purposefully not that

Solaxi:
Yes that makes totally sense. This was just an easier fix to implement than to remove something I really did not know if should or could be removed in the first place. :) Tried asking this in #development but since no one reacted, just thought that I'll make the change and y'all will reject it if not good.

But I don't think the check itself should be removed as it checks if private channels are setup correctly. Nor should we exclude the GM from this check either since GM can be also a player, right?
https://github.com/AsyncTI4/TI4_map_generator_bot/blob/58329c198ce34ff8408c2cd3cca14894eb57ccde/src/main/java/ti4/buttons/ButtonListener.java#L195

Fingon00:
The GM is a player, but not a real player. There should be an additional check after player != null that checks if the player is a real player (player.isRealPlayer()). That would avoid sending the GM bad warnings, and also avoid setting up a private channel for the GM, who isnt a real player

Issue #1346, opened 2024-01-16
Whisper edge cases


ninevra:
Whispering to faction aliases does not work. For example, `toempyrean` sends a message but `toempy` does not. Some of the whisper code supports aliases and some does not and overall it works out to not.

Whispering to a faction or color not in the game sends the whisper to yourself. This is only really a problem with `tofutureplayer` whispers. For those the sender is told they "sent someone else a future message" with no indication that anything is wrong until their next turn, at which point they receive the message and the log says they whispered to themselves.

Whispering to a faction or color + any suffix sends the whisper to yourself. For example `tosaarball hi` sends `hi` back to the sender. Again only a problem with `tofutureplayer` whispers.

jamesrudolph:
I had an idea for the empty whisper last night.

What if an empty whisper just reports back to the user:
"You tried to send an empty whisper. This whisper would have gone to XYZ player."
This would double as a quick tester for folks to be certain they are whispering the person they want to whisper.

PR #1776, opened 2024-05-19
Checks to tocolor messages

Solaxi:
- Notify sender if target player was not found, rather than sending the message to yourself as currently happens.
- Check in FoW is target player is a neighbour and notify if not.


jamesrudolph:
Fixes #1346 I think?

PR #1780, opened 2024-05-20
Franken draft mode to pick 1 at a time

Solaxi:
Our group thinks picking 3 from the first bag is unbalanced and we want to pick only 1 at a time.

Ideally would've like to make it a choice of how many to pick from first bag and how many from the following bags, but as we currently only save the draft name and not any options for it, felt this is more convenient solution at the moment.

Also don't really like that command has two options which only one can used (powred or onepick). More futureproof could've been /franken start_franken_draft mode: default/powered/onepick. But then tooltip about the mode woudn't be visible.

Thoughts?


jamesrudolph:
Use an enum for draft types instead of bools, then you can include the tooltip in the autocomplete. Let me find an example

jamesrudolph (approved):

jamesrudolph:
/statistics games
GameStats.java (not the bothelper one)
Has an example of an enum which is easy to add additional types. See AutoCompleteProvider.java for  how the enum is consumed and presents the description/tooltip to the end user

PR #1790, opened 2024-05-23
Support for multiple tiles for add/remove token

Solaxi:
As this already supports multiple players, added similar support for multiple tiles. Did a lot of testing, but since several commands extend this one and this affects all of them, I'd like a review to confirm this is ok to do. :)

Main benefit for custom maps to be able to add all custom wormholes at once, 
image
or to remove them
image

But works with any token.

Also allows adding or removing multiple ccs if there is a reason to
image


Fingon00:
The change seems simple enough, and I tested a few things of my own, should be good

PR #1801, opened 2024-05-26
Don't diplo a system for non-real players

Solaxi:
In FoW it spoils the supprise of an NPC player being present when CC is added for a color players know is not one of them.


jamesrudolph (approved):
Good opportunity to dedupe/refactor this so ButtonHelper calls a DiploSystem.diploSystem

Solaxi:
Good point. I'll look into it!

PR #1802, opened 2024-05-26
SetHomeSystemPosition command to set HS explicitly in Franken

Solaxi:
An override setting to explicitly set home system position in Franken. 

Default behaviour persists unless homeSystemPosition is set. In that case use that one.

image

This allows moving stats_anchors to better locations so player information won't overlap the map mainly in Franken-FoW where they can't be moved atm due to visibility checks based on HS. Currently an issue for example in my fow180.


jamesrudolph (commented):
I wouldn't only use this variable for franken, i.e. I wouldn't want a dev thinking player.getHomeSystemPosition is any different than player.getHomeSystemTile().getPosition()

I'd look at using the DataMigrationManager to do an initial seed of that variable using player.setHomeSystemTileLocation(player.getHomeSystemTile)

I'd also set this variable during player setup (same location stats anchor first gets set)

PR #1820, opened 2024-06-01
Fow options command for optional settings

Solaxi:
Added a command /fow fow_options to put all fow settings to one place and make it easy to add more if needed in the future.

- Moved the setting to hide player names from /custom customization to /fow fow_options
- Added an option to hide total votes in agenda
- Show set fow options in /game info if game is fow

### DIFF
--- a/src/main/java/ti4/map/GameSaveLoadManager.java
+++ b/src/main/java/ti4/map/GameSaveLoadManager.java
@@ -1800,7 +1800,7 @@ public class GameSaveLoadManager {
                         // Do nothing
                     }
                 }
-                case "fow_hide_names" -> { //REMOVE THIS AFTER ONE SAVE/LOAD GAMES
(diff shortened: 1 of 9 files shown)


src/main/java/ti4/map/GameSaveLoadManager.java
```diff
@@ -1787,10 +1800,11 @@ private static void readGameInfo(Game game, String data) {
                         // Do nothing
                     }
                 }
-                case Constants.FOW_OPTION_HIDE_NAMES -> {
+                case "fow_hide_names" -> { //REMOVE THIS AFTER ONE SAVE/LOAD GAMES
```
jamesrudolph, line 1803: Add "TODO" so it shows up as a warning in an IDE

jamesrudolph:
My only real qualm with using a HashMap instead of an explicit value for kind of thing is that it makes it harder for any other dev to know what is already stored within the hashmap.

E.g. a dev needs to know this option exists now instead of just browsing intellisense for game.


You're good to push this as is though

PR #1857, opened 2024-06-13
Update pbd2000.json

HolyTispoon:
adding text for raven faction (no longer a secret).


HolyTispoon:
not sure what's up with VSC but I guess I don't use it enough to keep it working.

PR #1870, opened 2024-06-28
(untested) Hexagonal Versions of Spiral Hexes

bradleysigma:
https://ti4.westaddisonheavyindustries.com/game/pbd3966 wanted to use the Creuss Home System as a Minor Faction, and did not want to put the Creuss Gate in the map with the Creuss Home System in `tr`. This creates a hexagonal version of the Creuss Home System that will fit in the map (as well as both sides of the wormhole nexus).


jamesrudolph:
image
I think it's because of the source not existing in the Source.java file

PR #1920, opened 2024-07-14
Add `/server_promote` command

bradleysigma:
This command should allow elevated users on the main server to promote themselves to an equivalent rank on the overflow servers.
- Admin, Developer and Bothelper can all promote themselves to Admin on the Emoji Farm severs;
- Admin can promote themselves to Admin, Developer and Bothelper on any of the regular servers;
- Developer can promote themselves to Developer and Bothelper on any of the regular servers;
- Bothelper can promote themselves to Bothelper on any of the regular servers;
- Everyone can also demote themselves from a rank they could promote themselves to.

This command only works in the `#development` channel, as a form of moderation.


bradleysigma:
Also, this is **untested**, given that it needs the multi-server architecture to be testable.

Jazzxhands (approved):

PR #1932, opened 2024-07-17
Silliness Moderation

bradleysigma:
When the Scepter of Emelpar is used, it will now be called by that name, **_mostly_**.

### DIFF
--- a/src/main/java/ti4/buttons/ButtonListener.java
+++ b/src/main/java/ti4/buttons/ButtonListener.java
@@ -763,8 +763,15 @@ public class ButtonListener extends ListenerAdapter {
                     }
                 }
             } else {
+                String scepterName = "Scepter of Emelpar";
+                switch (ThreadLocalRandom.current().nextInt(20))
+                {
+                    case 1: scepterName = "Scepter of Empelar"; break;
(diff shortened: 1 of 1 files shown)


Fingon00:
I sorta liked scepter of silly spelling ;_;

Fingon00:
Ya but this change is like, 85% less fun

bradleysigma:
¯\_(ツ)_/¯, I think it's better in small doses, for the same reason eyes only appear on the destroyers every so often, instead of all the time.

Fingon00:
Maybe we could just misspell it 19 times 🤔 and have silly spelling as the 20th. That'd be fun

Fingon00:
> ¯\_(ツ)_/¯, I think it's better in small doses, for the same reason eyes only appear on the destroyers every so often, instead of all the time.

It is already in small doses, being once a round and only present in some games.

bradleysigma:
Hmmmm, could have it just generate a random permutation for the middle letters?

Fingon00:
Would be funny, though I do bet we'd get a bunch of bug reports complaining about it being misspelled. Not sure I actually wanna deal with those, though the thought of the joke is amusing

PR #2023, opened 2024-08-10
Updates

bradleysigma:
This change will compress anything on a tile with Mirage towards the bottom left edge (except Mirage itself), which should prevent/reduce Mirage overlapping something important. It also moves the destroyer to the left on a `TYPE08` (blank) tile.  
Without Mirage:  
image  
With Mirage:  
image-1  
Corner case: Planet Stellar Converted then Empyrean Hero then Mirage:  
image


bradleysigma:
Inserted images are now dragged from their centres instead of their tops lefts:  
image  
Also, changes if Mirage appears in a multi-planet system:  
image

Issue #24, opened 2022-03-28
/show_game option for wide view for use on pc


HolyTispoon:
currently the image is so big that's really hard to see without zooming in.
The idea: add an optional variable to show_game that has options map/stats/full (defaults to full). Map and stats only show you that part of the image of ease of readability. BUT here’s the great part: when you update unit movements on the map the automatic message of the bot shows only the map so you could more easily confirm what happened. Same for stats.

aogden:
This browser extension https://github.com/aogden/TI4LandscapeMap is a bandaid for this open issue

bradleysigma:
`/show_game display_type: landscape` will now show a wide view, with the map on the left and the scores, objectives, laws and player areas on the right.

PR #2037, opened 2024-08-18
Add landscape map option

bradleysigma:
`/show_game display_type: landscape` will create a landscape map, with the map on the left and the stats on the right. Implements #24 .


bradleysigma:
pbd22_2024 08 18_-_09 12 20-1

PR #2160, opened 2024-10-02
Voices of the Council

Cryypter:
Adding in Voices of the Council homebrew game mode


src/main/resources/data/strategy_cards/cryypter.json
```diff
@@ -0,0 +1,12 @@
+[
+    {
+        "id": "cryypter_3",
+        "initiative": 3,
+        "name": "Politics",
+        "primaryTexts": [],
```
jamesrudolph, line 6: Ideally you would add the card's texts here, like:
[
"Clause1",
"Clause2",
"Clause2"
]

jamesrudolph:
@Cryypter do you have more to add or can this be pushed? I tested this on my local machine and the game setup appears to be working as expected

jamesrudolph (approved):

PR #2180, opened 2024-10-14
absol tech fixes 14oct24

scottmk:
- **fix: missing War Sun button on production for Muaat players on Absol's Techs & Mechs**
- **fix: incorrect tech requirement's for Absol's Mahact Crimson Legionnaire II**
- **fix: bug where Absol's AI Development Algorithm's post-pass tech research is paid for with Resources instead of Influence**
- **fix: bug where Absol's AI Development Algorithm post-pass ability provided buttons for all tech types instead of just units**

### DIFF
--- a/src/main/java/ti4/buttons/ButtonListener.java
+++ b/src/main/java/ti4/buttons/ButtonListener.java
@@ -1046,15 +1046,19 @@ public class ButtonListener extends ListenerAdapter {
             ButtonHelper.deleteMessage(event);
         } else if (buttonID.startsWith("getAllTechOfType_")) {
             String techType = buttonID.replace("getAllTechOfType_", "");
-            boolean noPay = false;
+            String payType = null;
             if (techType.contains("_")) {
-                techType = techType.split("_")[0];
-                noPay = true;
+                final String[] split = techType.split("_");
+                techType = split[0];
+                payType = split[1];
             }
             List<TechnologyModel> techs = Helper.getAllTechOfAType(game, techType, player);
-            List<Button> buttons = Helper.getTechButtons(techs, techType, player);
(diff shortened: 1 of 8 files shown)


src/main/java/ti4/buttons/ButtonListener.java
```diff
@@ -2800,6 +2805,7 @@ public void resolveButtonInteractionEvent(ButtonContext context) {
                     ButtonHelper.deleteMessage(event);
                 }
                 case "acquireATech" -> ButtonHelper.acquireATech(player, game, event, messageID, false);
+                case "acquireAUnitTechWithInf" -> ButtonHelper.acquireATech(player, game, event, messageID, false, Set.of(Constants.UNIT), "inf");
```
jamesrudolph, line 2808: We're trying to move away from adding to ButtonListener and using @ButtonHandler notation instead, although I do see a bunch of params here that won't get passed through the ButtonHandler. You can either leave it as is or create a wrapper method over ButtonHelper.acquireATech to pass the messageID, false, Set, & "inf" in

src/main/java/ti4/map/Player.java
```diff
@@ -543,7 +543,7 @@ public boolean hasInf2Tech() {// "dszeliinf"
     @JsonIgnore
     public boolean hasWarsunTech() {
```
jamesrudolph, line 544: Is this the method used to determine whether you're allowed to build a warsun?
Hm, can you see a way to just use the unitmodel for this? I.e. player.getUnitByBasetypeID("warsun") != Null? Future proofs any new warsun homebrew

jamesrudolph (approved):

PR #2189, opened 2024-10-18
Fow improvements

Solaxi:
* Added a check for existing spectator thread, so spectating space combat followed by invasion uses the same thread like it does for the actual combat thread. Currently it was creating a new one for invasion.

* Allow non-players to create EndOfRound summaries. So basically GM can use `endofround1 Message` to save the first summary. Buttons to add/modify them are given in reply so after the first one, they can use the buttons too. 
  * Used UserName instead of Faction for the key to save them
  * Used Good Doggos instead of Faction Emoji

* Allowed the use of `/combat combat_roll` in channels not starting with the game name. If it is used in a thread and the parent channel name is valid game, then it's allowed.
  * Helps a lot in FoW if player-to-player private threads can run that to do PDS throws.

### DIFF
--- a/src/main/java/ti4/helpers/async/RoundSummaryHelper.java
+++ b/src/main/java/ti4/helpers/async/RoundSummaryHelper.java
@@ -12,8 +12,9 @@ import net.dv8tion.jda.api.interactions.components.text.TextInput;
 import net.dv8tion.jda.api.interactions.components.text.TextInputStyle;
 import net.dv8tion.jda.api.interactions.modals.Modal;
 import net.dv8tion.jda.api.interactions.modals.ModalMapping;
+import net.dv8tion.jda.api.entities.channel.middleman.MessageChannel;
 import ti4.buttons.Buttons;
-import ti4.helpers.ButtonHelper;
+import ti4.helpers.Emojis;
 import ti4.helpers.RegexHelper;
 import ti4.listeners.annotations.ButtonHandler;
 import ti4.listeners.annotations.ModalHandler;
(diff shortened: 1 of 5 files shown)


Solaxi:
Added fix for /relic send_fragment which count option of `Number of fragments (default 1)` had changed to be 3.

src/main/java/ti4/helpers/async/RoundSummaryHelper.java
```diff
@@ -66,21 +68,31 @@ public static void finishEditRoundSummary(ModalInteractionEvent event, Game game
...
         game.setStoredValue(roundKey, previousThoughts + thoughts);
 
-        MessageHelper.sendMessageToChannelWithButton(player.getCardsInfoThread(), player.getFactionEmoji() + " stored an end of round summary", Buttons.EDIT_SUMMARIES);
+        MessageChannel playerChannel = player.isRealPlayer() ? player.getCardsInfoThread() : eventChannel;
+        MessageHelper.sendMessageToChannelWithButton(playerChannel, resolvePlayerEmoji(player) + " stored an end of round summary", Buttons.EDIT_SUMMARIES);
         MessageHelper.sendMessageToChannel(game.getMainGameChannel(), "Someone stored an end of round summary");
     }
+
+    public static String resolveRoundSummaryKey(Player player, String roundNum) {
+        String keySuffix = player.isRealPlayer() ? player.getFaction() : player.getUserName();
```
jamesrudolph, line 91: can you check if an eliminated player would get their end of round summary (after elimination) would get posted as a separate good doggo entry?

src/main/java/ti4/helpers/async/RoundSummaryHelper.java
```diff
@@ -66,21 +68,31 @@ public static void finishEditRoundSummary(ModalInteractionEvent event, Game game
...
+        String roundKey = resolveRoundSummaryKey(player, roundNum);
         String previousThoughts = "";
         if (append && !game.getStoredValue(roundKey).isEmpty()) {
             previousThoughts = game.getStoredValue(roundKey);
             previousThoughts = previousThoughts.replaceFirst("\\.? ?$", "") + "\n";
         }
         game.setStoredValue(roundKey, previousThoughts + thoughts);
 
-        MessageHelper.sendMessageToChannelWithButton(player.getCardsInfoThread(), player.getFactionEmoji() + " stored an end of round summary", Buttons.EDIT_SUMMARIES);
+        MessageChannel playerChannel = player.isRealPlayer() ? player.getCardsInfoThread() : eventChannel;
```
jamesrudolph, line 85: you can check, but player.getCardsInfoThread() _should_ still work for any not-real player

jamesrudolph (approved):

src/main/java/ti4/helpers/async/RoundSummaryHelper.java
Solaxi: Yes getCardsInfoThread will create a thread if one would not exist. Didn't want card thread to be made for GM in FoW. They have their GM channel (which we have no good way of getting other than searching by channel name) and this seemed the simplest way to get endofround summaries to work from there.

src/main/java/ti4/helpers/async/RoundSummaryHelper.java
Solaxi: Will do. I have time tomorrow. :)

PR #2284, opened 2024-11-08
Unit Rendering Refactor

heisenbugged:


src/main/java/ti4/generator/UnitRenderGenerator.java
```diff
@@ -0,0 +1,545 @@
...
+        boolean isSpace = unitHolder.getName().equals(Constants.SPACE);
+        boolean isMirage = unitHolder.getName().equals(Constants.MIRAGE);
+        boolean hasMirage = false;
+        if (isSpace) {
+            Set<String> tokenList = unitHolder.getTokenList();
+            hasMirage = tokenList.stream().anyMatch(tok -> tok.contains("mirage"))
+                && (tile.getPlanetUnitHolders().size() != 3 + 1);
+        }
+
+        Map<String, String> jailTiles = Map.of(
```
jamesrudolph, line 242: We can remove the "jailTiles" handling - these are deprecated and have been replaced the NomBox in the player area tied to the /capture commands

src/main/java/ti4/generator/UnitRenderGenerator.java
```diff
@@ -0,0 +1,545 @@
...
+    private BufferedImage getUnitDecal(Player player, UnitKey unitKey) {
+        if (player == null) return null;
+        return ImageHelper.read(resourceHelper.getDecalFile(player.getDecalFile(unitKey.asyncID())));
+    }
+
+    private BufferedImage getSpoopyImage(UnitKey unitKey, Player player) {
+        BufferedImage spoopy = null;
+
+        // Random spoopy warsun
+        if (unitKey.getUnitType() == UnitType.Warsun && ThreadLocalRandom.current().nextInt(1000) == 0) {
```
jamesrudolph, line 390: just because I see it, use CalendarHelper.isNearHalloween() to control whether spooky warsun chance is 1/10 or 1/1000

jamesrudolph:
To check:
- all special units
- unit decals
- mirage in a 3 planet system
- systems with 3 or more player's units in them
- unit tags

src/main/java/ti4/generator/UnitRenderGenerator.java
```diff
@@ -0,0 +1,545 @@
...
+            case Fighter -> unitPath.replace(Constants.COLOR_FF, Constants.BULK_FF);
+            case Infantry -> unitPath.replace(Constants.COLOR_GF, Constants.BULK_GF);
+            default -> unitPath;
+        };
+
+        Player player = game.getPlayerByColorID(unitKey.getColorID()).orElse(null);
+        if (player == null) return unitPath;
+
+        // Handle special unit cases
+        if (unitKey.getUnitType() == UnitType.Spacedock && player.ownsUnitSubstring("cabal_spacedock")) {
```
jamesrudolph, line 364: Not sure if this was finished yet, but there was a project to stop using the "csd" image for these and instead be painting the rift under the normal space dock image instead, to eliminate the need for graphics folks to create the csd version for each colour.
If that doesn't look to be handled yet, then we can leave this as is

src/main/java/ti4/generator/UnitRenderGenerator.java
```diff
@@ -0,0 +1,545 @@
...
+        return calculateSpiralPosition(posCtx, radius, degree, degreeChange, rectangles);
+    }
+
+    private Point getTokenPosition(UnitKey unitKey, PositioningContext posCtx) {
+        boolean fighterOrInfantry = Set.of(UnitType.Infantry, UnitType.Fighter).contains(unitKey.getUnitType());
+        String positionKey = fighterOrInfantry ? "tkn_" + posCtx.unitId : posCtx.unitId;
+        return ctx.unitTokenPosition.getPosition(positionKey);
+    }
+
+    private Point calculateSpiralPosition(
```
jamesrudolph, line 462: I'll attempt a draw of all prod games and check to see if we enter this method at all

jamesrudolph:
Some special units are shifting inwards

New:
image

Old:
image

jamesrudolph:
yeah, special units are being stacked:

New:
image

Old:
image

src/main/java/ti4/generator/TileGenerator.java
```diff
@@ -74,7 +73,7 @@ public TileGenerator(Game game, GenericInteractionCreateEvent event) {
     }
 
     public TileGenerator(Game game, @Nullable GenericInteractionCreateEvent event, @Nullable DisplayType displayType,
-                         int context, String focusTile) {
```
wholton, line 77: We should keep these spaced over to avoid them being on the same line as the code within the method.

src/main/java/ti4/generator/TileGenerator.java
```diff
@@ -86,8 +85,8 @@ public TileGenerator(Game game, @Nullable GenericInteractionCreateEvent event, @
 
     private boolean isFowModeActive() {
         return game.isFowMode() && event != null &&
-                (event.getMessageChannel().getName().endsWith(Constants.PRIVATE_CHANNEL) ||
-                        event instanceof ShowGameAsPlayer.SlashCommandCustomUserWrapper);
+            (event.getMessageChannel().getName().endsWith(Constants.PRIVATE_CHANNEL) ||
```
wholton, line 88: This class moved to `src/main/java/ti4/image`, you should be able to move both these classes over, take your changes (if it has merge conflicts), then compare with master to make sure no unintended changes were missed.

src/main/java/ti4/generator/TileGenerator.java
heisenbugged: i think this was just automatically done by the vscode lint settings

PR #2303, opened 2024-11-12
Change emoji on custodians message

teuschb:
Not sure I got the emoji name right


teuschb:
Can someone check the emoji name before I push this?

PR #2502, opened 2024-12-05
Update Planet Emoji

bradleysigma:
Update to some higher quality images for some of the planet emoji.


Jazzxhands:
Discord will resize these down to 128x128 anyway, it would be better to just do it here preemptively. Otherwise looks good to me

bradleysigma:
Will do.

bradleysigma:
Hmmmm, just realised you said 128 and not 256, but I can't be bothered to make the changes right now; just let the bot handle the resizing, I guess.

PR #2537, opened 2024-12-09
Replace complete with queue

SunshinePoncho:
Replaced all .complete() with .queue() hoping that's all I did but I will admit my cat climbs on my keyboard more than id admit in public


src/main/java/ti4/buttons/UnfiledButtonHandlers.java
```diff
@@ -1498,7 +1498,7 @@ public static void deleteButtons(ButtonInteractionEvent event, String buttonID,
             if ("Done Producing Units".equalsIgnoreCase(buttonLabel)) {
                 MessageHistory mHistory = event.getChannel().getHistory();
                 RestAction<List<Message>> lis = mHistory.retrievePast(3);
-                Message previousM = lis.complete().get(1);
+                Message previousM = lis.queue().get(1);
```
wholton, line 1501: It'll require a bit more than this. The point of moving from complete() to queue() is that you can provide queue() a callback, something that consumes the eventual result of the operation and let's the current thread move forward and do its own thing. On the otherhand, complete() requires the running thread to wait for the completion of the task before moving forward.

wholton (changes requested):
Will try to give pointed advice later, but for now do not merge.

PR #2625, opened 2024-12-20
Filter out spacedocks for assault cannon combat

### LINKED ISSUE #2613: Saar Space Factory should not be targetable by Assault Cannon technology
Saar is able to select their faction unique Space Dock tech Space Factory I/II when selecting their target for an opponent's Assault Cannon.  Since their Space Factories are considered ships only for moving and retreating, they are not targetable during space combat and should not be selectable.

Wezabi:
Fixes #2613 

Map
Target List 1
Target List 2


wholton:
@Wezabi Looks good. Thanks.

PR #2636, opened 2024-12-22
Change Start-Of-Turn Message

bradleysigma:
Currently, the bot says `[player] is UP NEXT` at the start of their turn. However, on the map, the active player is marked `ACTIVE` and the next player in initiative order is marked `NEXT UP`. This change primarily changes the former message to be more explicit that it is the player's turn *now*.

The player who is next in initiative order is also shown in the start of turn message (if not fog-of-war), but they are not pinged. The message changes if all other players have passed. In addition, the buttons for doing the turn action are moved below the pending strategy cards etc, so that the message can be resolved top-to-bottom (though in the example below, the strategy cards were popped in initiative order, so ideally Leadership would be up the top).

image

### DIFF
--- a/src/main/java/ti4/commands2/player/SCPick.java
+++ b/src/main/java/ti4/commands2/player/SCPick.java
@@ -256,11 +256,12 @@ class SCPick extends GameStateSubcommand {
                 String text = player.getRepresentationUnfogged() + ", it is now your turn (your " 
                     + StringHelper.ordinal(player.getInRoundTurnCount()) + " turn of round " + game.getRound() + ").";
                 Player nextPlayer = EndTurnService.findNextUnpassedPlayer(game, player);
-                if (nextPlayer == null || game.isFowMode()) {
-                } else if (nextPlayer == player) {
-                    text += "\n-# All other players are passed; you will take consecutive turns until you pass, ending the action phase.";
-                } else {
-                    text += "\n-# " + nextPlayer.getRepresentationNoPing() + " will start their turn once you've ended yours.";
+                if (nextPlayer != null && !game.isFowMode()) {
+                    if (nextPlayer == player) {
+                        text += "\n-# All other players are passed; you will take consecutive turns until you pass, ending the action phase.";
+                    } else {
+                        text += "\n-# " + nextPlayer.getRepresentationNoPing() + " will start their turn once you've ended yours.";
+                    }
                 }
                 MessageHelper.sendMessageToChannel(game.getMainGameChannel(), text);
                 if (privatePlayer.getGenSynthesisInfantry() > 0) {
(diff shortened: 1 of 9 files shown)


src/main/java/ti4/commands2/player/SCPick.java
```diff
@@ -234,21 +238,31 @@ public static void secondHalfOfSCPick(GenericInteractionCreateEvent event, Playe
...
                 privatePlayer.setInRoundTurnCount(privatePlayer.getInRoundTurnCount() + 1);
                 if (game.isShowBanners()) {
                     BannerGenerator.drawFactionBanner(privatePlayer);
                 }
-                MessageHelper.sendMessageToChannelWithButtons(game.getMainGameChannel(), "\n Use Buttons to do turn.",
-                    StartTurnService.getStartOfTurnButtons(privatePlayer, game, false, event));
+                String text = player.getRepresentationUnfogged() + ", it is now your turn (your " 
+                    + StringHelper.ordinal(player.getInRoundTurnCount()) + " turn of round " + game.getRound() + ").";
+                Player nextPlayer = EndTurnService.findNextUnpassedPlayer(game, player);
+                if (nextPlayer == null || game.isFowMode()) {
```
wholton, line 259: It's a code-smell to have empty if statements like this. I'd change it to:

```            if (nextPlayer != null && !game.isFowMode()) {
                if (nextPlayer == player) {
                    text += "\n-# All other players are passed; you will take consecutive turns until you pass, ending the action phase.";
                } else {
                    text += "\n-# " + nextPlayer.getRepresentationNoPing() + " will start their turn once you've ended yours.";
                }
            }```

Comment (bradleysigma) [reply src/main/java/ti4/commands2/player/SCPick.java]: Fixed.

File: src/main/java/ti4/commands2/player/SCPick.java [hunk 2]
```diff
@@ -216,7 +220,7 @@ public static void secondHalfOfSCPick(GenericInteractionCreateEvent event, Playe
                 if (game.isShowBanners()) {
                     BannerGenerator.drawFactionBanner(privatePlayer);
                 }
-                MessageHelper.sendMessageToChannelWithButtons(privatePlayer.getPrivateChannel(), msgExtra + "\n Use Buttons to do turn.",
+                MessageHelper.sendMessageToChannelWithButtons(privatePlayer.getPrivateChannel(), "Use buttons to do turn.",
```
Solaxi, line 223: Why msgExtra removed here? It's set in 207 but then not used here.

bradleysigma: I could be wrong, but I think that `msgExtra` is sent to the player on (new) line 212 with `MessageHelper.sendPrivateMessageToPlayer`, and so it was getting sent twice?

PR #2703, opened 2024-12-31
Templating: Misc

bradleysigma:


src/main/java/ti4/helpers/ButtonHelperAbilities.java
```diff
@@ -1073,7 +1073,7 @@ public static void pillageCheck(Player player, Game game) {
                 + playerIdent
                 + ". Please check this is a valid **Pillage** opportunity, and use buttons to resolve.";
             buttons.add(Buttons.red(finChecker + "pillage_" + player.getColor() + "_unchecked",
-                "Pillage " + playerIdent));
+                "Pillage " + player.getFlexibleDisplayName()));
```
bradleysigma, line 1076: @wholton, I think this should fix the PILLAGE issue.

Issue #2699, opened 2024-12-30
Duranium Armor used on destroyed unit


Wezabi:
If a unit with sustained damage is destroyed, the auto-assign message should not include it being repaired.  

![1000007254](https://github.com/user-attachments/assets/ab5bbc61-f260-40bb-a55c-8513a68bed0a)

Wezabi:
I can provide more info if needed.  The bot does not try to repair the destroyed unit when you select "auto-assign hits" so that's working correctly.  This is just to fix the message.

PR #2721, opened 2025-01-03
Duranium Armor summary fix & repair order automation

Wezabi:
# Overview
Fix for #2699

Issue was in `#autoAssignSpaceCombatHits` when `justSummarizing == true` , destroyed units are not tracked before reaching final duranium check. 

- Added a map to track damaged units instead
- Added sort by preferred repair order (opposite hit order assignment).

I went pretty overboard on test cases, it was just faster for me to test personally. I am open to whatever feedback and best practices the devs prefer.  

# Example Combat with Changes
### DA on highest value ship
Screenshot 2025-01-03 115045
### DA on highest value ship with damage
Screenshot 2025-01-03 115308
### DA ship when hits can be assigned to lower value ship
Screenshot 2025-01-03 115403
### DA not used as ship was undamaged
Screenshot 2025-01-03 115446
### DA not used as ship was blown up
Screenshot 2025-01-03 115959

### DIFF
--- a/src/main/java/ti4/helpers/ButtonHelperModifyUnits.java
+++ b/src/main/java/ti4/helpers/ButtonHelperModifyUnits.java
@@ -6,6 +6,7 @@ import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.TreeMap;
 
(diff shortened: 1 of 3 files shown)


Fingon00:
Pretty cool stuff

PR #2722, opened 2025-01-03
Get replacement mid-milty draft working again

michaelryanmcguire:
MiltyDraftManager was throwing a null pointer when it tried to loop over the players using its updated copy of userIds because we never updated the Game objects Map of players by id to have the new userId in it.


Jazzxhands (approved):
Looks good!

PR #2792, opened 2025-01-20
Ignis aurora consolidation

jamesrudolph:


adam133:
@jamesrudolph Good enough for now - I'd like to get it to print all errors before failing but the right syntax for setting/getting a variable is avoiding me.

PR #2793, opened 2025-01-20
add a check for invalid json files

adam133:


jamesrudolph:
See https://github.com/AsyncTI4/TI4_map_generator_bot/pull/2792 where I pulled this in to test it

@adam133 Any way to have this actually fail the test, and run it before the RunTests (probably a requires: on RunTests?)

adam133:
> See #2792 where I pulled this in to test it
> 
> @adam133 Any way to have this actually fail the test, and run it before the RunTests (probably a requires: on RunTests?)

Yeah I can tweak that real quick - looking

jamesrudolph:
merged with https://github.com/AsyncTI4/TI4_map_generator_bot/pull/2792

PR #2806, opened 2025-01-24
Filter out colored Promissory Notes from search [skip ci]

Jazzxhands:


jamesrudolph:
Remove this, that I had just added: https://github.com/AsyncTI4/TI4_map_generator_bot/blob/master/src%2Fmain%2Fjava%2Fti4%2Fautocomplete%2FAutoCompleteProvider.java#L789

PR #2913, opened 2025-03-02
force remove existing lock file if one exists

adam133:


adam133:
Gonna merge this, feel like it should be safe enough.

Issue #2489, opened 2024-12-04
Add "save" as a field in the ButtonHandler annotation with a default of true


adam133:
Modify the code to check for that when deciding to save in the ButtonContext. Then, go through various ButtonHandlers and turn off saves if they're just info buttons. Bonus for moving the various ButtonHandlers you modify to the Button directory into related classes.

Additional context from Jazz:
> When we initially enumerate all the button handlers, we create a Consumer<ButtonContext> which ultimately calls the tagged function. How this would look is that we would have an extra shouldSave property on ButtonContext as well, and inside that consumer we'd flip it based on the handler
> So technically the context doesn't look at the handler, the handler modifies the context

Lezduit:
Someone doing this already? looks like a decent issue for getting up to speed with the repo, can i pick it up?

Jazzxhands:
@Lezduit this one is a pretty low-level change, though I'd expect it to be pretty small overall. So, yeah go for it! If you need help or pointers feel free to reach out :)

PR #2940, opened 2025-03-12
Save trigger refactor

Lezduit:
Some extra attention for the removed static list of non-save buttons which will be removed from ButtonContext in this change, i couldn't find all of those buttonIDs with a ButtonHandler or didn't know how to transform them from the case/switch to the ButtonHandler framework.


Lezduit:
@Jazzxhands we dont use github at work, how do i link this to the related issue? or do we just not?

Jazzxhands:
You can just #2489

Fingon00:
Why are you not saving on bestowTitleStep2? Do we not currently save on that step?

Lezduit:
@Fingon00 We currently do save at that step, and the one before.
@Jazzxhands explained we should save at anything that modifies the game, i assumed this is not game related.
If i have been overzealous in disabling saves, let me know where you want m enabled and i'll adjust.

Jazzxhands:
> @Fingon00 We currently do save at that step, and the one before.
> @Jazzxhands explained we should save at anything that modifies the game, i assumed this is not game related.
> If i have been overzealous in disabling saves, let me know where you want m enabled and i'll adjust.
Yah, this one is a bit overzealous because titles are stored in the game object that they came from. You can see in the body of the method it does `game.setStoredValue(...)` which we do want to save

Fingon00:
From experience today @Jazzxhands : users will expect to be able to undo any button press that changes things for them, even ones that dont change the game state. If we stop saving at those points, then when those users go to UNDO, they will undo things they dont intend to undo and that will break things, potentially making this a fix that causes many more problems than it fixes

src/main/java/ti4/helpers/PlayerTitleHelper.java
```diff
@@ -69,7 +69,7 @@ public static void resolveBestowTitleStep1(Game game, Player player, ButtonInter
         ButtonHelper.deleteMessage(event);
     }
 
-    @ButtonHandler("bestowTitleStep2_")
+    @ButtonHandler(value = "bestowTitleStep2_", save = false)
```
wholton, line 72: because this calls "game.set...", it must save..

wholton:
```
if (titles.isEmpty()) {
            game.setStoredValue("TitlesFor" + p2.getUserID(), title);
        } else {
            game.setStoredValue("TitlesFor" + p2.getUserID(), titles + "_" + title);
        }
```

src/main/java/ti4/helpers/PlayerPreferenceHelper.java
```diff
@@ -198,7 +198,7 @@ public static List<Button> getDirectHitManagementButtons(Game game, Player playe
         return buttons;
     }
 
-    @ButtonHandler("riskDirectHit_")
+    @ButtonHandler(value = "riskDirectHit_", save = false)
     public static void resolveRiskDirectHit(Game game, Player player, ButtonInteractionEvent event, String buttonID) {
```
wholton, line 202: `game.setStoredValue("stuffNotToSustainFor" + player.getFaction(), stuffNotToSustain);`

wholton: This has to save.

src/main/java/ti4/helpers/TransactionHelper.java
```diff
@@ -720,7 +720,7 @@ private static boolean sendMemeInsteadOfText(ButtonInteractionEvent event, Game
         return RandomHelper.isOneInX(1000);
     }
 
-    @ButtonHandler("sendOffer_")
+    @ButtonHandler(value = "sendOffer_", save = false)
```
wholton, line 723: `game.setStoredValue(key, offerNumber + "");`

wholton (changes requested):
This is a great change, but go through and double check the spots you have modified to make sure they do not make any changes to the Game object, or its underlying objects like Player.

If the method itself is already doing a save, it's fine. Also, any "UserSetting" changes are also fine to ignore.

Thanks.

Jazzxhands:
@Fingon00 it's the users who are wrong 😆

Lezduit:
@wholton thanks for taking the time to review, modified based on your comments.

wholton (approved):

PR #2991, opened 2025-03-29
initialize new environment emoji cache

ClaysGit:
Create emojis.json from discord when working in a new environment.

In new environments, your application will have most icons already but your cache doesn't exist. This creates the cache for you, instead of failing to upload 1000+ icons.


Jazzxhands (dismissed):
If not too much trouble, please make sure your formatter is working :)

Otherwise looks good!

ClaysGit:
Thanks for the call out on the formatter @Jazzxhands

Issue #2880, opened 2025-02-18
Reorganize BotLogger


Jazzxhands:
- Rewrite BotLogger.java to have 3 main methods: `info(...)`, `warning(...)`, and `error(...)`
- Update all usages in the code to use one of these three methods 
- Hook up info to post to #bot-log-info, warning to #bot-log-warning, and error to #bot-log-error
  - if those channels can't be found, look for #bot-log, and then finally resort to the webhook

wholton:
On point 2, it's okay to slowly move things over.

PR #3012, opened 2025-04-07
Logging Rework (issue #2880)

JustOneOther:
## Purpose

This PR implements and expands upon the features requested in #2880.

## Feature list

- BotLogger.info(...), BotLogger.warn(...), and BotLogger.error(...) have been fully implemented and have the following changes:
    - All methods now look for a dedicated channel before sending to the universal #bot-log channel.
    - All methods use accept `BotLogger.LogMessageOrigin` to trace data about their source for more accurate logging.
    - All methods now have a consistent format including a timestamp, source trace (if available), event data (if available), stack trace (if available), and required manual message.
    - All methods now attempt to respect the discord rate limit (see Known Issues for details)
- Slash Command logs and Button Command logs have the following changes:
    - Both logs will now search for a dedicated channel before defaulting to creating a thread in the primary guild's bot log channel.
    - Both logs now contain a `BotLogger.LogMessageOrigin` to trace data about their source for more accurate logging.
    - Both logs now cache their logs via `InteractionLogCron` before sending them to reduce repeated websocket/webhook calls.
- The `BotLogger.LogMessageOrigin` class has been introduced to store and retrieve data about logging messages' sources.
- The `InteractionLogCron` class has been introduced to cache and bulk-send command logs, reducing the number of websocket/webhook calls.
- All other methods of logging not listed above have been deprecated.
- All existing logging calls have been converted to a method listed above.

## Patch List

- `DiscordWebhook`'s JSON implementation no longer causes malformed payloads when given strings with newlines.

## Known Issues

- Discord may rate limit log messages, returning a 429 error. Fixing this requires a substantial rewrite of the `DiscordWebhook` implementation, and as such was deemed out of scope for this PR. This issue will not affect normal functionality of the bot, as the rate limit is resource-specific to the bot-log channel webhook.

## Local Tests

The PR passed all tests on a local instance (93 passed, 1 ignored). It also demonstrated complete functionality of features listed in Feature List (see above).

## Notes

Commits made roughly adhering to Conventional Commits.
This is a minor version bump (SemVer), deprecating but not removing methods.


src/main/java/ti4/AsyncTI4DiscordBot.java
```diff
@@ -296,7 +299,7 @@ private static void startBot(Guild guild) {
         CommandListUpdateAction commands = guild.updateCommands();
         CommandManager.getCommands().forEach(command -> command.register(commands));
         commands.queue();
-        BotLogger.logWithTimestamp(" BOT STARTED UP: " + guild.getName());
+        BotLogger.info(new BotLogger.LogMessageOrigin(guild), "BOT STARTED UP: " + guild.getName());
```
wholton, line 302: I would maybe just pass in "guild" rather than "new BotLogger.LogMessageOrigin(guild)", then do the latter inside the BotLogger itself.

src/main/java/ti4/autocomplete/AutoCompleteProvider.java
```diff
@@ -71,7 +71,7 @@ public static void handleAutoCompleteEvent(CommandAutoCompleteInteractionEvent e
         try {
             resolveAutoCompleteEvent(event);
         } catch (Exception e) {
-            BotLogger.log("Error in handleAutoCompleteEvent", e);
+            BotLogger.error(new BotLogger.LogMessageOrigin(event), "Error in handleAutoCompleteEvent", e);
```
wholton, line 74: Same here, I would consider just passing in "event".

src/main/java/ti4/cron/InteractionLogCron.java
```diff
@@ -0,0 +1,92 @@
...
+
+public class InteractionLogCron {
+	@Nonnull
+	private static final ArrayDeque<BotLogger.AbstractEventLog> messageCache = new ArrayDeque<>(500);
+	@Nonnull
+	private static final HashMap<Class<?>, TextChannel> logChannels = new HashMap<>();
+	private static TextChannel primaryBotLogChannel;
+	@Nonnull
+	private static final HashMap<Class<?>, StringBuilder> messageBuilders = new HashMap<>(); // This prevents needing to rebuild the hashmap every time logInteractions is called, and is memory-cheap as it holds references
+	private static boolean isRegistered = false;
```
wholton, line 24: `private static boolean isRegistered;`

src/main/java/ti4/cron/InteractionLogCron.java
```diff
@@ -0,0 +1,92 @@
...
+import java.util.concurrent.TimeUnit;
+
+public class InteractionLogCron {
+	@Nonnull
+	private static final ArrayDeque<BotLogger.AbstractEventLog> messageCache = new ArrayDeque<>(500);
+	@Nonnull
+	private static final HashMap<Class<?>, TextChannel> logChannels = new HashMap<>();
+	private static TextChannel primaryBotLogChannel;
+	@Nonnull
+	private static final HashMap<Class<?>, StringBuilder> messageBuilders = new HashMap<>(); // This prevents needing to rebuild the hashmap every time logInteractions is called, and is memory-cheap as it holds references
```
wholton, line 23: `private static final Map<Class<?>, StringBuilder> messageBuilders = new HashMap<>();`

wholton: Also, I'd put the comment above rather than on the same line, since it's a bit long.

src/main/java/ti4/message/BotLogger.java
```diff
@@ -12,264 +20,815 @@
...
-        return false;
-    }
+
+	private static long lastScheduledWebhook = 0;
+	public static final long DISCORD_RATE_LIMIT = 50; // Min time in millis between discord webhook messages
+
+	/**
+	 * Describes the discord-based origin of a log message and handles fetching data for log messages from these sources.
+	 */
+	public static class LogMessageOrigin {
```
wholton, line 41: I'm tempted to say put this in its own file, for readability since it's quite large. They're tightly coupled, so I get the point, but there's a trade-off.

If it stays in this class, put it at the very bottom of the file.

src/main/java/ti4/message/BotLogger.java
```diff
@@ -12,264 +20,815 @@
...
+			this.originTime = DateTimeHelper.getCurrentTimestamp();
+		}
+
+		public LogMessageOrigin(@Nonnull GenericInteractionCreateEvent event) {
+			this.event = event;
+			if (event.isFromGuild()) {
+				this.channel = event.getGuildChannel();
+				this.guild = event.getGuild();
+			} else
+				BotLogger.warning("LocationSource created from non-guild event. This will not attribute messages.");
```
wholton, line 82: To match up with the code elsewhere, prefer to add braces around if/else, unless it's all on one line.

src/main/java/ti4/message/BotLogger.java
```diff
@@ -12,264 +20,815 @@
...
+			this.originTime = DateTimeHelper.getCurrentTimestamp();
+		}
+
+		public LogMessageOrigin(@Nonnull Game game) {
+			this.game = game;
```
wholton, line 87: No need to specify `this`, unless you're just being super-extra cautious in case someone adds extra parameters to the method or something.

src/main/java/ti4/message/BotLogger.java
```diff
@@ -12,264 +20,815 @@
...
+					.stream()
+					.findFirst()
+					.orElse(null));
+		}
+	}
+
+	/**
+	 * Enum for data associated with log severity
+	 */
+	public enum LogSeverity {
```
wholton, line 248: Can this be package scoped, or private?

src/main/java/ti4/message/BotLogger.java
```diff
@@ -12,264 +20,815 @@
...
+	 */
+	private static void scheduleWebhookMessage(@Nonnull String message) {
+		long timeToNextMessage = lastScheduledWebhook + DISCORD_RATE_LIMIT - System.currentTimeMillis();
+		if (timeToNextMessage <= 0) {
+			sendMessageToBotLogWebhook(message);
+		} else {
+			CronManager.scheduleOnce(MessageHelper.class, () -> sendMessageToBotLogWebhook(message), timeToNextMessage, TimeUnit.MILLISECONDS);
+		}
+
+		lastScheduledWebhook = System.currentTimeMillis() + Math.max(timeToNextMessage, 0);
```
wholton, line 505: Is this fine to not be thread safe?

src/main/java/ti4/cron/InteractionLogCron.java
```diff
@@ -0,0 +1,92 @@
...
+
+import javax.annotation.Nonnull;
+import java.util.ArrayDeque;
+import java.util.HashMap;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+public class InteractionLogCron {
+	@Nonnull
+	private static final ArrayDeque<BotLogger.AbstractEventLog> messageCache = new ArrayDeque<>(500);
```
wholton, line 18: Is this actually a cache? If I'm reading the code correctly, it came across as just a buffer of messages you're pushing every few minutes.

src/main/java/ti4/cron/InteractionLogCron.java
wholton: Since we're talking a 2 minute schedule, does it matter that it gets rebuilt?

src/main/java/ti4/message/BotLogger.java
JustOneOther: I'm going to leave it in BotLogger for the reasons you specified above, and have moved it to the bottom of the file along with the other subclasses and enums.

src/main/java/ti4/image/DrawingUtil.java
```diff
@@ -195,7 +195,11 @@ public static void drawCCOfPlayer(Graphics graphics, String ccID, int x, int y,
             if (player == null) gameName = "Null Player";
             if (player != null && player.getGame() == null) gameName = "Null Game";
             if (player != null && player.getGame() != null) gameName = player.getGame().getName();
-            BotLogger.log("Ignored error during map generation for `" + gameName + "`", e);
+            if (player != null) {
+                BotLogger.error(new BotLogger.LogMessageOrigin(player), "Ignored error during map generation for `" + gameName + "`", e);
+            } else {
+                BotLogger.error(new BotLogger.LogMessageOrigin(player), "Ignored error during map generation for `" + gameName + "`", e);
```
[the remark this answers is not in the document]
JustOneOther: Fixed, this was probably an artifact of blind method signature alteration.

src/main/java/ti4/cron/InteractionLogCron.java
JustOneOther: Moved into its respective method, this should be a lot cleaner. Thanks!

src/main/java/ti4/cron/InteractionLogCron.java
```diff
@@ -0,0 +1,92 @@
...
+	@Nonnull
+	private static final HashMap<Class<?>, TextChannel> logChannels = new HashMap<>();
```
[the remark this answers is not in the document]
JustOneOther: Fixed, thank you!

src/main/java/ti4/cron/InteractionLogCron.java
JustOneOther: Fixed, thank you!

src/main/java/ti4/message/BotLogger.java
JustOneOther: Fixed, thank you!

src/main/java/ti4/message/BotLogger.java
JustOneOther: This is an old habit of mine, mostly to be as explicit as possible. I'm going to leave it as-is if that's alright.

wholton: Sorry, you *should* specify `this` on this line, it's the other lines. Misclicked.

src/main/java/ti4/message/BotLogger.java
JustOneOther: This is an artifact of an earlier design. It is now set to private.

src/main/java/ti4/cron/InteractionLogCron.java
wholton: Mind you, this is more about *connotation* of cache vs buffer. It's maybe fine as-is, but I know when I first glanced at it I didn't understand what it was going to be doing.

JustOneOther: Definitely a misnomer here, this has been renamed to `messageBuffer`.

src/main/java/ti4/message/BotLogger.java
wholton: And, yes, that's fine.

src/main/java/ti4/message/BotLogger.java
JustOneOther: There is one issue here, which is passing `MessageHelper.class` instead of `BotLogger.class`, but the lambda itself should be thread-safe as Strings are immutable, and all other passed variables are local and not reused or final.
Please note that I have very little experience threading in java, so I may be missing some java-specific difference here.

wholton: I meant that `lastScheduledWebhook` is a static field that might be changed by multiple threads coming into this method at the same time.

JustOneOther: Ah gotcha. I didn't think about logging methods being called from different threads. I'll update that to be synchronized.

JustOneOther: This should now be fixed via a dummy lock.

wholton (dismissed):

JustOneOther:
Merging this pr is on hold pending a bug fix

wholton (approved):

PR #3015, opened 2025-04-10
hotfix: Default to searching for primary bot-log channel before moving to webhook

JustOneOther:
Hotfix for #3012 as specified in title.

### DIFF
--- a/src/main/java/ti4/message/BotLogger.java
+++ b/src/main/java/ti4/message/BotLogger.java
@@ -238,7 +238,7 @@ public class BotLogger {
 		String compiledMessage = msg.toString();
 		int msgLength = compiledMessage.length();
 
-		if (channel == null) {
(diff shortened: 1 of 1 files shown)


Jazzxhands (approved):

src/main/java/ti4/message/BotLogger.java
```diff
@@ -237,24 +238,33 @@ private static void logToChannel(@Nullable LogMessageOrigin origin, @Nonnull Str
 		String compiledMessage = msg.toString();
 		int msgLength = compiledMessage.length();
 
+		if (channel == null && AsyncTI4DiscordBot.guildPrimary != null) {
```
Jazzxhands, line 241: Probably also verify guildPrimary is not null here, for the tests

JustOneOther: Yep, just hotfixed it. Thanks!

PR #3022, opened 2025-04-12
Update ds.json

bjsensenig2:
Made iconoclasm faction ability and ATS Armaments non-draftable. Provided alternative text to Edyn hero to clarify that it checks for number of mechs instead of sigils


bjsensenig2:
Pull request has passed checks and is ready to be pushed

adam133 (approved):

PR #3037, opened 2025-04-19
Update ds.json

bjsensenig2:
Made Celestial Guides Non-draftable, so it does not show up in 2 places in the draft


bjsensenig2:
Simple Change all checks are clear

adam133 (approved):

PR #3051, opened 2025-04-28
Defer on creating default priority track

ClaysGit:
The Priority Track is created:
- When you enable the homebrew (too early, seat order may not be set correctly)
- When you proceed to the first strategy phase

This removes the former, as it screws up and the latter is already sufficient.

Also, I don't think I know how to manage commit history in a forked repo


ClaysGit:
Also added:
- Fix a lot more spots that use some kind of "turn order" to do the right thing in Omega Phase
- Added some new helpers to facilitate the above; would appreciate a close look at those since I'm less familiar with Java types
- Milty Drafting after enabling the homebrew should succeed, though the lack of explicit support means the agenda deck will be goofed up. Will continue to work on.

src/main/java/ti4/helpers/Helper.java
```diff
@@ -428,50 +440,41 @@ public static Player getPlayerWithThisSC(Game game, int sc) {
...
-        for (Player p2 : game.getRealPlayers()) {
-            if (p2 == player) {
-                found = true;
-                players.add(player);
-            } else {
-                if (found) {
-                    players.add(p2);
-                }
-            }
+    public static List<Player> getNonInitiativeOrder(Game game) {
```
Fingon00, line 443: Non initiative order is a bit of a confusing term. getSpeakerOrPriorityOrder is longer but at least more obvious what it's talking about

PR #3056, opened 2025-04-29
Fix Kyro Agent (Kyro gets to place the inf, not the target player)

mew1033:


Fingon00:
I have this change wrapped up in some other changes I'm pushin tonight

PR #3093, opened 2025-05-10
Fix display of secret objective scoring ping

Wezabi:
Status phase scoring reminder shows public objective message twice instead of players needing to score secrets. Fixed to use soMsg instead of poMsg for both cases.


Wezabi:
I didn't test this change but I assume it fixes the issue. If you want an example of a game where this issue occurred, lookup pbd9189

Fingon00:
Thanks for spotting this, I've wrapped this change into another PR to reduce bot down time.

PR #3117, opened 2025-05-18
When the bot automatically marks you as not following, it also adds a message to the thread, in addition to emoji react

derektumolo:


derektumolo:
expecting plenty of feedback, but i think this works well, and would definitely make things a bit clearer in the thread.

src/main/java/ti4/service/strategycard/PlayStrategyCardService.java
```diff
@@ -443,9 +442,44 @@ private static void sendAndHandleMessageResponse(MessageCreateData toSend, Game
...
+            && (!p2.getTechs().contains("iihq") || !p2.getUnfollowedSCs().contains(8))
+            && !p2.hasRelicReady("absol_emelpar") && !p2.hasRelicReady("emelpar")
+            && !p2.hasUnexhaustedLeader("mahactagent") && !p2.hasUnexhaustedLeader("yssarilagent")) {
+            return true;
+        } 
+        return false;
+    }
+
+    private static boolean playerCantFollowNoDocks(Game game, Player player,  Player p2, int scToPlay) {
+        if (scToPlay == 6 && !p2.hasUnit("ghoti_flagship") && !ButtonHelper.getTilesOfPlayersSpecificUnits(game, p2, UnitType.Spacedock).contains(p2.getHomeSystemTile())) {
```
Jazzxhands, line 479: Preexisting, but you could add an `&& !p2.getUnfollowedSCs().contains(4)` here, in case the player wants to get a dock from construction and then follow warfare

src/main/java/ti4/helpers/AgendaHelper.java
```diff
@@ -681,7 +681,7 @@ public static void pingNonresponders(Game game, String buttonID, ButtonInteracti
 
     }
 
-    private static String pluralPerson(int num) {
+    private static String pluralPerson(int num){
```
Jazzxhands, line 684: Not sure which IDE you're using, but make sure you configure the formatter. If you're using VSCode, it's just making sure the following two lines are in `.vscode/settings.json`:
```json
    "java.format.settings.url": "eclipse-formatter.xml",
    "editor.formatOnSave": true,
```

Jazzxhands (approved):
Neither comment is a big deal, really. Looks good

Fingon00:
Can you also do this for when it marks you as not following due to your pre-set no-follow?

PR #3138, opened 2025-05-25
Mapper cleaning + 2 new commands /search

schluik:
organised Mapper + refactoring of some function
migrated /search sources get...Sources() functions to Mapper added /search tokens and /search attachments
modified TokenModel and AttachmentModel
organised search values in Constants
also a few fix to data

### DIFF
--- a/src/main/java/ti4/autocomplete/AutoCompleteProvider.java
+++ b/src/main/java/ti4/autocomplete/AutoCompleteProvider.java
@@ -14,13 +14,14 @@ import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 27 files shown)


Fingon00:
Note that I believe /search is approaching the maximum number of subcommands. I would probably avoid adding more or migrate some existing ones out first

schluik:
> Note that I believe /search is approaching the maximum number of subcommands. I would probably avoid adding more or migrate some existing ones out first

I didn't know there was a limit, good to know,
I have a couple more I want to do, and then I should be done, I hope I won't reach the limit

schluik:
Failed some tests
Issue found and solved already
Second PR incoming

PR #3221, opened 2025-06-16
Spring boot

wholton:


adam133 (approved):

wholton:
@adam133 afaik, it was working prior to the recent push for Player Info. If we want to get this in, I want to push another API related to refreshing the map that I got working locally. Just haven't touched it since I haven't heard much interest.

src/main/java/ti4/spring/controller/ActionCardController.java
```diff
@@ -0,0 +1,57 @@
...
+        UserIdValidator.validate(gameName, userId);
+
+        discordOAuthService.authorize(authorizationHeader, userId);
+
+        Game game = GameManager.getManagedGame(gameName).getGame();
+        Player player = game.getPlayer(userId);
+        List<String> hand = actionCardService.getHand(player);
+
+        return new GetPlayerActionCards(hand);
+    }
```
heisenbugged, line 55: thoughts on a single endpoint for _all_  player specific secret information in a game? (which would be action cards but also secrets). otherwise we'd be doing two api calls on every (authenticated) game load when it could be just one

PR #3236, opened 2025-06-21
Queue franken draft bags

ninevra:
Still working on this, but most of the functionality is there.
## Goal

Pass bags as soon as players are done with them, and allow them to queue up for the next player.

### Game state

`Player#currentDraftBag` is now a queue of draft bags, the 0th of which is the current bag. Serializes as a list separated by semicolons. `Player#readyToPassBag` is removed.

### Game behaviors

The core logic is moved from `BagDraft#passBags()` to a trio of functions, `void passBag(Player)`, `DraftBag dequeueBag(Player)`, and `void enqueueBag(Player, DraftBag)`, which call eachother as necessary. A lot of the message-printing moves from the `FrankenButtonHandler`s into those functions as well, since the process is a bit more complicated now.

### UI

Current draft looks something like:
image

Indicates how many bags are with each player.

Would it be helpful to display some information about each bag? An arbitrary name, or a count of remaining items?

### Other stuff

The existing concept of "draft queues", referring to the items a player is in the process of choosing from their current bag, is renamed to "selections" throughout. This is for the sake of being able to read the code >_<

## TODO

Open questions:
- [ ] How to handle migration
    - Converting in-progress drafts to queueing requires very little code (just check at loading to see if a player is `readyToPassBag` and if so, pass it), but is confusing.
    - Allowing drafts to continue in both modes requires a lot of extra code/work, but is possible.

Remaining work:
- [ ] Handle the error/edge-case where a bag exists with no draftables for anyone.
- [ ] Test with more than 2 players. How to set up such a test game?
- [ ] Clean up code a bit.
- [ ] Automated tests? Maybe?


ninevra:
Experimenting with custom emojis

image

ninevra:
Implementing a choice of queueing or non-queueing drafts is complicated a little by the presence of the various `BagDraft` subclasses (`FrankenDraft`, `OnePickFrankenDraft`, etc.) The existing subclasses seem to represent draft policies, e.g. component category limits. Maybe move to composition for that (a `BagDraft` has a `DraftPolicy` which may be a `FrankenDraftPolicy` etc..) and use inheritance for `QueueingBagDraft` and `NonqueueingBagDraft`?

Fingon00:
Appreciate the work, and if you ever feel like picking it up again Ill work with you to get it in, but atm this is out of date and I dont think anyone wants to pick it up, so we'll table it for now

PR #3304, opened 2025-07-16
Update luminous.json

bjsensenig2:
Updated text for all of Luminous Strategy Cards


bjsensenig2:
I will note that part of the Luminous_1 or SC1 anticipation had the ready Agent, Relic, Technology to only apply to component you control, to avoid refreshing Zeu and agents of other players and other shenanigans

PR #3364, opened 2025-07-25
Update byz_agendas.json

byzTI:


jamesrudolph:
@byzTI do you still want these changes? noticing they've been sitting here for a bit

PR #3419, opened 2025-07-28
Test with Static Mocks

heisenbugged:


src/test/java/ti4/buttons/handlers/relics/RelicButtonHandlerTest.java
```diff
@@ -0,0 +1,78 @@
...
+            assertFalse(actualButtons.isEmpty());
+
+            boolean hasTacticalAction = actualButtons.stream()
+                .anyMatch(button -> button.getLabel().contains("End Turn"));
+
+            boolean hasComponentAction = actualButtons.stream()
+                .anyMatch(button -> button.getLabel().contains("Transaction"));
+
+            assertTrue(hasTacticalAction, "Should contain an end turn button");
+            assertTrue(hasComponentAction, "Should contain a transaction button");
```
heisenbugged, line 75: this violates single responsibility principle"
as long as we ensure the message helper is called that's sufficient

but i am just keeping it to show its possible

PR #3420, opened 2025-07-28
Service Registry Tests

heisenbugged:


src/main/java/ti4/service/ServiceRegistry.java
```diff
@@ -0,0 +1,38 @@
...
+    private static RelicService relicService;
+    @Setter
+    @Getter
+    private static ComponentActionService componentActionService;
+    @Setter
+    @Getter
+    private static MessageService messageService;
+    @Setter
+    @Getter
+    private static DynamicRelicButtonHandler dynamicRelicButtonHandler;
```
heisenbugged, line 22: (this works but ... just use spring lol)

PR #3422, opened 2025-07-28
Test with Spring DI

heisenbugged:


src/main/java/ti4/service/ComponentActionService.java
```diff
@@ -0,0 +1,19 @@
...
+
+/**
+ * Service for component action operations to enable dependency injection and testing
+ */
+@Service
+public class ComponentActionService {
+
+    public void serveNextComponentActionButtons(ButtonInteractionEvent event, Game game, Player player) {
+        ComponentActionHelper.serveNextComponentActionButtons(event, game, player);
+    }
```
heisenbugged, line 18: if we go with this we can just migrate the actual logic. this shim layer is just done for prototype simplicity

PR #3429, opened 2025-07-29
Run thread archival check on a cron

wholton:
## Summary
- add `ThreadArchiveCron` scheduled every minute
- register cron in `AsyncTI4DiscordBot`

## Testing
- `mvn -q test` *(fails: Non-resolvable parent POM)*

------
https://chatgpt.com/codex/tasks/task_e_6888c56be3e4832d9b13e00fab7fd1ca


Fingon00:
Isnt it better to run the check when new threads are going to be created? To avoid the possibility of >20 threads being created in under a minute being an issue?

PR #3506, opened 2025-08-15
Refactor tactical action service

heisenbugged:
This refactors tactical actions to have colocated responsibilities

The core tactical action code is moved to `TacticalActionDisplacementService` .. which does all things around handling the displacement map representing movement

All the "button assembly" code to let users trigger their abilities have been grouped into three categories

`movement abilities` -- things you trigger during general movement
`post movement abilities` -- things you trigger once your movement is done from a system
`planet abilities` (few) -- things you trigger on planets 

this follows a similar pattern to the agenda resolver, using a registry instead of just interminable if/else with shared interfaces


heisenbugged:
@jrkd sticking with the existing terminology mostly

jrkd:
@heisenbugged no worries. Ive been away from the code for a while so keeping current terms sounds right to me. 

Also, Ive made a bunch inline comments - feel free to ignore or disagree. I havent pulled it cause i dont have a local java setup anymore and better to also get someone whos currently active like Will have a go through too but saw the review come through on discord and thought i could offer a once over

src/main/java/ti4/service/tactical/movement/MoveAbilities.java
```diff
@@ -0,0 +1,285 @@
...
+import net.dv8tion.jda.api.interactions.components.buttons.Button;
+import ti4.buttons.Buttons;
+import ti4.helpers.ButtonHelper;
+import ti4.helpers.FoWHelper;
+import ti4.helpers.Units.UnitType;
+import ti4.service.emoji.ExploreEmojis;
+import ti4.service.emoji.FactionEmojis;
+import ti4.service.emoji.SourceEmojis;
+
+public final class MoveAbilities {
```
wholton, line 13: This smells like an enum to me.

heisenbugged:
@jrkd 
thanks! rather than respond to every individual comment high level

**On DI**
100% agree, but there are some differing team opinions on this. I actually have an [open pr](https://github.com/AsyncTI4/TI4_map_generator_bot/pull/3422) showing spring DI integration but haven't merged it yet, so totally on the same page!

**On abilities having button code**
Your review has made it clear to my my naming is bad. All the "abilities" here aren't ability _execution_ rather 
_ability button registering_, so it's all intrinsically tied to discord. (I don't love the `contribute()` method but side-point). I will rename to make sure the word `Button` is included so that's clearer.

**On ability refactoring**
You made some comments about the internals of the ability, most of which I agree with, but I'm mostly not touching the actual implementation details here and simply porting them over (don't want to introduce bugs)


This is "round 1" and there are several rounds to come, continuing to separate more. The Web API *will* need to trigger discord buttons as it will only handle the movement part of a tactical action, so exactly what that boundary looks like I leave to future implementation


thank you so much!

PR #3758, opened 2025-10-07
Count X89c4 for player stats area

ClaysGit:
Without X-89, Mentak's starting infantry average 1.2 hits. With X-89, it's 2.4 hits. This PR ensures the player area shows that upgrade correctly:
image

### DIFF
--- a/src/main/java/ti4/helpers/ButtonHelper.java
+++ b/src/main/java/ti4/helpers/ButtonHelper.java
@@ -2360,7 +2360,7 @@ public class ButtonHelper {
                         hitChance = 1 - ((1 - hitChance) * (1 - hitChance));
                     }
                     float combatValue = removedUnit.getBombardDieCount() * hitChance * uh.getUnitCount(unit);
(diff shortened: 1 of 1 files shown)


wholton:
Can you do this for bombardment as well?

```
if (removedUnit.getBombardDieCount() > 0) {
                    hitChance = ((11.0f - removedUnit.getBombardHitsOn()) / 10);
                    if (game.playerHasLeaderUnlockedOrAlliance(player, "jolnarcommander")) {
                        hitChance = 1 - ((1 - hitChance) * (1 - hitChance));
                    }
                    count += removedUnit.getBombardDieCount() * hitChance * uh.getUnitCount(unit);
                }
```

Also, you don't have to do this, but it'd be nice to move these helper functions out of the ButtonHelper class... as they have nothing to do with buttons.

ClaysGit:
Added bombardment, though I'm not sure where it's used; can't visually confirm I did it right.

I agree that these should be refactored out, but not just the functions I'm touching...and this branch is not intended for a larger refactor.

PR #3806, opened 2025-10-20
NewUI Colour Display Name

jamesrudolph:
Add metadata for the proper colour name, vs displaying the ID
image


jamesrudolph:
@heisenbugged for adding stuff to newui, is it easiest/best to just push the data through this way, or on the web side, use the component ID to get it from the datamodels? i.e. send ColorDisplayName directly, or for web to lookup the ColorDisplayName from the ColorModel data via ColorID?

PR #3819, opened 2025-10-23
fix: corrected Twilight's Fall abilities to match US English version of cards

scottmk:
Some of the cards were totally incorrect, and many of them had slight variations from the actual card texts.

I referenced the physical cards when making these edits.

NB: Two of the cards had incorrect names that necessitated alias changes, and one of them had an incorrect faction


scottmk:
I believe this PR is more comprehensive than https://github.com/AsyncTI4/TI4_map_generator_bot/pull/3817

PR #3834, opened 2025-10-25
Franken map build

ClaysGit:
Adds the `/franken build_map` command. Can be used once the franken draft has finished.

Required a lot of re-writing of the MantisMapBuildService to work without the draft system manager.

I've also discovered null analysis settings, so some changes are due to that.


src/main/java/ti4/website/AsyncTi4WebsiteHelper.java
ClaysGit: These were clogging the output pane where I see all the errors I'm throwing, so I figure hide them by default @heisenbugged

PR #4038, opened 2025-11-26
Veiled heart/publicly show the type of each veiled card

JayCyrZak:
Veiled cards are facedown, but not fully hidden, which means that their types (ability/unit/genome/paradigm) are visible to all players.


JayCyrZak:
bug: veiled genomes/paradigms don't show up if the player has no unveiled ones
fixing this real quick on a different branch

PR #4635, opened 2026-04-09
remove api gateway callout, unused

adam133:


wholton:
You can undo the changes you made to the test class. I fixed it. Was failing due to the color mapper not being setup. Need this class to extend a parent class that handles that setup.

adam133:
@copilot can you revert this commit 88eae14720c18926941d1f7d19790d286a9030d4

PR #4731, opened 2026-04-17
fix: prevent sardakk agent slash exhaust crash

heisenbugged:
## Summary
- handle the `/leaders exhaust` Sardakk agent slash-command path when no button-encoded planet target is present
- route players back into the existing tactical-action button flow when an active system with planets is available
- avoid the `ArrayIndexOutOfBoundsException` reported in Rollbar item 19 (occurrence 458589330477)

## Test Plan
- `mvn spotless:apply`
- `DB_PATH=./src/test/resources RESOURCE_PATH=./src/main/resources mvn --batch-mode --update-snapshots --no-transfer-progress verify test-compile`


Fingon00:
Ideally what I want to do someday with this is have a button in cards info for (almost) every agent which goes through the same flow as exhausting the agent with a slash command does -- it prompts the player to select the user of the agent and then directs them to proper resolution. RN this is only functional for some agents (and some agents it will never be suitable for). 

This maybe my project today though, to get several more agents on this path.

PR #4738, opened 2026-04-18
Update pok.json

bjsensenig2:
Sardakk BT Balance Update


bjsensenig2:
Yeah Idk why I am failing this test, what is the best way for me to understand this check failure?

jamesrudolph:
Hm, apparently it's the existence of "fragile" in the draft able items, but I don't see why "fragile" shows up here but not on other branches.

PR #4781, opened 2026-04-22
Harden startup lease handoff

heisenbugged:
## Summary
- move startup orchestration into explicit phases in `main()` and run lease-owned startup work through a named callback
- harden lease claiming/renewal/release with dedicated transactional operations and single-winner conditional updates
- abort startup when JDA or primary guild initialization fails instead of continuing into readiness and lease participation

## Test Plan
- `mvn -q -DskipTests compile`

### DIFF
--- a/src/main/java/ti4/AsyncTI4DiscordBot.java
+++ b/src/main/java/ti4/AsyncTI4DiscordBot.java
@@ -1,5 +1,8 @@
 package ti4;
 
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
 import org.springframework.context.ConfigurableApplicationContext;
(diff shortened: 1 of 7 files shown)


src/main/java/ti4/spring/service/jda/JdaLifecycleService.java
```diff
@@ -12,21 +11,16 @@
...
-        JdaService.initialize(resolveSourceArgs());
-    }
-
     @PreDestroy
     private void shutdown() {
         JdaService.shutdown();
     }
 
-    private String[] resolveSourceArgs() {
+    public String[] resolveSourceArgs() {
```
wholton: Move this to where it's used

src/main/java/ti4/AsyncTI4DiscordBot.java
```diff
@@ -18,6 +24,32 @@ public static void main(String[] args) {
...
+        }
+        JdaService.loadStaticDataAndResources();
+        JdaService.indexGameNames();
+        activeLeaseService.beginLeaseParticipation(AsyncTI4DiscordBot::runLeaseOwnedStartupWork);
+        JdaService.registerAndStartCronJobs();
+        JdaService.markProcessReady();
+    }
+
+    static void runLeaseOwnedStartupWork() {
+        if (DataMigrationManager.runMigrations()) {
```
wholton, line 48: The migrations use the managed games. We used to not accept commands until they were finished, as far as I remember. We probably want to keep that.

Basically, the startup process should freeze if there is a migration to run, waiting for the warmup to finish. No commands should run until they are done.

heisenbugged: Yes, that's how it works. There's two steps. Step one is acquiring the lease, step two is being 'ready'.

"Process Ready" is really what unleashes the floodgates. 

So we acquire the lease FIRST (so old instance cannot do work), run migrations, and THEN process ready.

PR #4824, opened 2026-04-24
Add combat replay side bets

heisenbugged:
## Why
Add configurable side bets to combat replay contests so users can spend leaderboard points on tracked combat outcomes during the voting window.

## What changed
- Adds side-bet persistence, button handling, placement validation, and replay-thread announcements.
- Tracks side-bet outcomes as top-level candidate fields from structured combat/action-card data.
- Resolves side-bet payouts into leaderboard scoring with per-type rewards.
- Simplifies combat contest settings to startup-only settings and removes the old enum constraint migration cleanup path.

## Validation
- mvn spotless:apply
- mvn spotless:check -Dspotless.check.skip=false
- mvn test -Dtest=CombatReplayServiceTest -Dspotless.check.skip=true
- git diff --check


heisenbugged:
Closing and reopening from the main repository branch instead of fork.

PR #4885, opened 2026-04-28
Add --fail-at-end to maven verify

jamesrudolph:


jamesrudolph:
@copilot I'd like the full test suite to run when the run-tests file itself is edited

PR #5046, opened 2026-05-11
Fix Franken draft apply with missing cards-info thread

heisenbugged:
## Summary
- Apply Franken draft bags only to real players, matching the rest of the bag draft flow.
- Resolve each player cards-info thread once and skip/log that player if the thread is unavailable instead of throwing in `MessageV2Builder`.

## Test Plan
- Not run, per request.


heisenbugged:
Closing as redundant: current master already includes the missing cards-info thread handling for Franken draft apply via #5040. The remaining diff here is only logging style, so this PR no longer needs to be rebased.

PR #5052, opened 2026-05-11
Fix missing cards-info thread recovery

heisenbugged:
## Summary
- Fixes a Cards Info button failure when a player's saved cards-info thread ID points at a Discord channel that no longer exists.
- Discord returns `10003 Unknown Channel` for that stale ID; the old flow let that exception abort before trying the existing thread-name fallback.
- The fix treats that specific response as a stale stored pointer, clears the saved thread ID, and then continues through the normal thread-name lookup/new-thread creation path.

## Test Plan
- Not run, per request.


wholton:
Made some changes to fix what was causing this -- FoW threads are in their own private channels, not in the main game channel

PR #5057, opened 2026-05-12
Recover stale cards-info thread ids

heisenbugged:
## Summary
- Recover when a saved cards-info thread id returns Discord 10003 Unknown Channel.
- Clear the stale id and continue through the existing name lookup/new-thread creation path.
- Add a reusable Discord unknown-channel helper alongside the existing unknown-message/webhook helpers.

## Test Plan
- mvn spotless:check
- git diff --check
- Not run per request: tests
- Attempted compile-only verification with `mvn -DskipTests compile`; blocked by local JDK 21 because the project requires release 26.

### DIFF
--- a/src/main/java/ti4/game/Player.java
+++ b/src/main/java/ti4/game/Player.java
@@ -1,5 +1,7 @@
 package ti4.game;
 
+import static ti4.helpers.discord.DiscordHelper.isUnknownChannelError;
+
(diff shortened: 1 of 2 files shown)


heisenbugged:
Superseded by #5052, which solves the same stale cards-info thread id recovery path.

PR #5058, opened 2026-05-12
Fix agenda autoresolve context recovery

heisenbugged:
## Summary
- detect plain manual autoresolve presses when there is no active agenda resolution context
- reply ephemerally that the agenda resolution window has closed and manual handling is needed if players still want the effect
- prevent stale agenda resolution buttons from reaching agenda button generation with incomplete currentAgendaInfo

## Verification
- mvn spotless:apply
- git diff --check
- mvn -DskipTests compile (blocked locally: project requires release 26, local JDK is 21.0.4)


wholton:
I don't know enough about this flow to say, but if you feel comfortable with it, then it's cool to go.

heisenbugged:
@wholton its just a friendlier message if people press agenda resolution buttons after agenda has passed - there's quite a bit of that in the error logs

PR #5065, opened 2026-05-12
Fix stale managed game channels

heisenbugged:
## Summary
The stack trace shows a guild member join event calling `RoleService.checkIfNewUserIsInExistingGamesAndAutoAddRole`, which sends “has joined the server” to the game table talk channel through `MessageHelper.sendMessageToChannel`. Discord rejected that send with `10003: Unknown Channel`.

The likely cause is that `ManagedGame` was holding a cached JDA channel object from warmup or a previous save. If that Discord channel was deleted or disappeared from JDA, the bot could still try to send through the old object and hit Unknown Channel.

This change keeps the existing cached channel fields, but when `ManagedGame` returns a table talk, actions, or launch post thread channel, it first looks up that same channel ID in the current JDA cache. If JDA no longer has the channel, the getter returns `null` instead of returning a stale channel object.

## Test Plan
- `JAVA_HOME=$(/usr/libexec/java_home -v 26) mvn -q -DskipTests compile`
- `git diff --check`

Tests were not run per request.

### DIFF
--- a/src/main/java/ti4/game/persistence/ManagedGame.java
+++ b/src/main/java/ti4/game/persistence/ManagedGame.java
@@ -13,6 +13,7 @@ import net.dv8tion.jda.api.entities.Guild;
 import net.dv8tion.jda.api.entities.channel.concrete.TextChannel;
 import net.dv8tion.jda.api.entities.channel.concrete.ThreadChannel;
 import org.apache.commons.lang3.StringUtils;
+import ti4.discord.JdaService;
 import ti4.game.Game;
 
 @Getter
@@ -38,7 +39,6 @@ public class ManagedGame {
     private final long endedDate;
     private final int round;
     private final Guild guild;
-    private final TextChannel mainGameChannel;
     private final TextChannel actionsChannel;
     private final TextChannel tableTalkChannel;
     private final ThreadChannel launchPostThread;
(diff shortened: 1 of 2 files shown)


wholton:
Hm... ManagedGames are recreated on every save.

`gameNameToManagedGame.put(game.getName(), new ManagedGame(game));`

heisenbugged:
@wholton well, [it's happening here](https://discord.com/channels/943410040369479690/1341438476528123904/1503747376332931144)

wholton:
> @wholton well, [it's happening here](https://discord.com/channels/943410040369479690/1341438476528123904/1503747376332931144)

Hm. I am gonna make changes along these lines, but my biggest concern is *why* it is happening. It's ultimately failing in `checkIfNewUserIsInExistingGamesAndAutoAddRole`, which sends a message to the table talk channel. How would that become stale? Seems freaky to me. If I have a Channel object saved in my ManagedGame... it exists?

wholton:
Gonna close this one. I don't think it resolves the underyling reason we are getting unknown channels. Once a managed game has a channel set, it shouldn't change. 

>  If that Discord channel was deleted or disappeared from JDA, the bot could still try to send through the old object and hit Unknown Channel.

Particularly this.

PR #5161, opened 2026-05-29
Franken Mahact Y-set errata

IslandForShans:
This adds Luminous' Mahact franken-only rework to franken modes.

This replaces the normal Mahact counterparts for franken-only.
The franken-only check is done via the franken-errata JSON.

Due to commander unlock being different for Franken version, I had to slightly change the commander unlock service. shouldBeUnlocked now checks wether it is "mahact" before going through the rest of logic. If it is "mahact" then there is special logic for it, if not it goes through the normal logic.

Mainly lots of additions to files with ability gates/message text. The actual mahact automation is untouched, just routed differently for the franken components.

### DIFF
--- a/src/main/resources/data/abilities/franken.json
+++ b/src/main/resources/data/abilities/franken.json
@@ -1,0 +1,24 @@
+[
+    {
+        "id": "primacy",
+        "name": "Primacy",
+        "faction": "mahact",
+        "window": "When you win a combat",
+        "windowEffect": "Place 1 owner token from your opponent on this ability if it does not already contain 1 of that player's tokens.",
+        "source": "franken"
+    },
+    {
+        "id": "edict_y",
+        "name": "Edict Y",
+        "faction": "mahact",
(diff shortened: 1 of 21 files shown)


src/main/resources/data/abilities/franken.json
```diff
@@ -0,0 +1,24 @@
+[
+    {
+        "id": "primacy",
+        "name": "Primacy",
+        "faction": "mahact",
+        "window": "When you win a combat",
+        "windowEffect": "Place 1 owner token from your opponent on this ability if it does not already contain 1 of that player's tokens.",
```
wholton, line 7: owner token is the same as control token?

wholton (approved):

src/main/resources/data/abilities/franken.json
IslandForShans: Correct, the wording is just slightly different. Feels more Mahact-like tbh 😂

PR #5209, opened 2026-06-06
Update pok.json

bjsensenig2:
Sardakk BT + Nomad Flagship Updates


bjsensenig2:
hmm idk why this isn't clearing the checks

IslandForShans:
If I had to guess the Franken errata test file is getting told that fragile is draftable when it expects it not to be?

PR #5252, opened 2026-06-19
Lore branch

Stabar-ti:
Significant addition to the Lore system -> While leaving the old functionality completely intact.

- On request made lore completely available in normal games
- Created automation on triggers

3 new files:
- 2 test files
- 1 Lore effects 

Most important double check might be player.java -> required to get a proper hook. Pretty sure I guarded it properly.

- Almost completely isolated from the rest of the main bot game files 
- Inserted minimal hooks and code into files outside of the lore code
- created tests against regression

* removal and placement of units 
* ac
* cc
* so
* vp
* tg / coms
* map modification: tile swap
* add and removal of tokens and attachments

Took into account:
* NPC replacement / neutral units
* replacements 

--UPDATE--

killed 3 bugs and 3 more edge cases. 

Also remove player.java change as the firing of it might be very obscure and more thought should be put into it.

--UPDATE2--

Removed an old persistancy bug from lore always being created to any game. 
Fow always has it turned one
Normal games now require setting it in weird game mode
Even when weird game mode lore is not toggled true, Lore can be precreated. A warning is hosted to toggle it.


src/main/java/ti4/game/Player.java
Fingon00, line 1: Game is initialized with round = 1. So checking if round > 0 is useless

Stabar-ti: I was not sure about that so I guarded it, I'll remove the check.

Stabar-ti: I reviewed it more. the only moment a pre firing could happen is when a home planet is placed with lore attached to it (this is very much an edge case. I'll log it in potential things to fix in the future and remove the additions to player.java the rest of the player.java changes I would like to keep as it is integral for one of the automation triggers.

Stabar-ti:
Just let me know on discord if you think other changes are needed!

PR #5298, opened 2026-06-27
Add random game setup command

olneyce:
## Summary
Adds `/milty random_setup` as an alternative to the normal Milty draft flow.

The command opens the existing Milty setup menu with the usual relevant setup options, then creates a randomized Milty-style setup that skips player drafting.

## Details
- Adds a new `/milty random_setup` subcommand.
- Reuses the existing Milty settings flow where possible.
- Hides slice/faction count settings because random setup uses the player count.
- Randomly assigns players, factions, slices, and speaker order.
- Adds a very light sanity check to reroll extreme outcomes where a player gets a weak slice, late speaker position, and weak faction together.
- Automatically resolves Keleres flavor selection during random setup.

## Testing
- Ran formatter successfully.
- Ran compile successfully.


olneyce:
Local build passed with `mvn -DskipTests package`.

This branch now incorporates the extracted `MiltyRandomSetupService` structure from #5302, while keeping the tested Random Galactic Events and True Random Galaxy options.

PR #5322, opened 2026-06-30
Update byz_agendas.json

byzTI:
Updating some of the agendas to see if this works


Fingon00:
Incorporating it into the other PR (as ya should have done)

PR #5362, opened 2026-07-07
Refactor transaction item parsing

heisenbugged:
## Summary
- Add a small TransactionItem value object for the existing stored transaction item format.
- Use it for transaction matching, clearing, display, public event sanitizing, and trade offer rendering.
- Keep saved transaction item serialization exactly unchanged.

## Test Plan
- mvn -Dtest=ti4.service.transaction.TransactionItemTest,ti4.helpers.TransactionHelperTest test


Fingon00:
I find the flow of this much harder to parse/work with. It will make it more difficult to debug/fix/add to transactions in the future as a result.

PR #5373, opened 2026-07-09
Disable combat replay by default

heisenbugged:
## Summary
- set the combat contest dev config to disabled by default
- prod config is already disabled on latest master

## Testing
- verified both combat contest config files report `enabled: false`


heisenbugged:
Closing: latest master already has prod combat replay disabled, and dev should remain enabled. The correction branch now has no file changes.

PR #5384, opened 2026-07-11
Refactor board and planet queries

heisenbugged:
## Summary

- centralize planet lookup, ownership, board traversal, and unit queries
- replace repeated planet-selection loops with reusable controlled/usable planet services
- consolidate agenda exhaustion, Smothering Presence, unit movement, and repeated faction workflows
- remove obsolete duplicate helpers and migrate call sites to canonical APIs

## Why

Common board and planet operations were repeatedly implemented through nested tile, holder, planet, and player loops. This made rules harder to read and encouraged duplicate helper APIs. The new domain-level queries keep rule predicates visible while removing storage-hierarchy boilerplate.

## Scope

Combat and combat-replay implementation files and tests are intentionally excluded from this PR. Deprecated compatibility delegates remain temporarily so unchanged combat code continues to compile against the new core APIs.

## Validation

- `mvn test`
- `git diff --check`
- staged-only snapshot tested with combat/replay changes excluded


Fingon00:
I like the naming of getRepresentation more than toString, given how sometimes we want to get representation without a ping or fogged/unfogged.

Fingon00:
This change is also hard to review because so many of the changes are just changing getRepresentation into toString. I dont have time to sift through all of those to find the real stuff.

PR #5385, opened 2026-07-11
Add player-scoped unit value injection

IslandForShans:
Adds a service that can inject values into any unit. If no argument exists it is a no-op and returns.

Tested with a homebrew ability. It gives move 1 to Mechs and it succesfully gave that value to a mech from a different faction only when that player had that ability.


IslandForShans:
@Fingon00 is this decided to be good to go? I can do more testing if you're wanting something specific.

IslandForShans:
Ok I ran some more tests. Nothing broken. Only thing is once it is injected into the unit the game treats that unit as having that value. This is intended of course but can lead to issues if the wrong value is injected.

PR #5464, opened 2026-07-27
Refactor token planet tile render system

IslandForShans:
This changes the rendering on tiles for the token planets to follow a slot-based point system for up to 6 token planets in one system. 

Created (but not adding) multiple different regression tests for all different kinds of scenarios involving token planets. All returned green.

Testing in live test bot also returned no issues.

Only about 3 will ever be in one system (more with ridiculous homebrew setups) so it should never reach overflow. 

Thunders edge remains in the center of the tile regardless of other present token planets. 

Avernus placement/movement causes recalculation of token planet positions in the tile.

If there is only one token planet in a system it defaults to the original token planet placement system.

Added support for 4 homebrew token planets.


IslandForShans:
Oh also any in progress games are functionally unaffected. The next time the map is updated with these token planets in a game it'll render them. This is all purely visual changes. Ran this by Meme Philosopher to make sure doing this wouldn't screw up the new ui.

PR #5517, opened 2026-08-10
Fracture rework - API, FOG bugs and handling, isFracture to enable token handling

Stabar-ti:
# Fracture: make it a tile/token property, and fix it in Fog of War

-> REASON

Fracture space had no data flag. `TileBack.FRACTURE` is never read anywhere in Java, so the real check was `position.startsWith("frac")` in ~30 places. That hardcodes the seven `frac1`–`frac7` slots, which rules out a fracture space token, and made "is the Fracture already here?" a guess.

-> WHAT IS CHANGED:

- **`isFracture` on `TileModel`** (optional, same pattern as `isNebula`), set on `fracture1`–`fracture7`. `Tile.isFracture()` = token → flag → legacy `frac` position, so nothing that used to be Fracture stops being Fracture. All ~30 gameplay checks now use it; rendering/slot checks stay positional.
- **Fracture space token** — `token_fracture_async.png` registered, so `/add_token fracture` makes any system Fracture space.
- **Can't place it twice** — `isFractureInPlay` is now "any tile with `isFracture()`", tiles *and* tokens. `spawnFracture` returns whether it placed anything; callers gate ingress on that. Also fixes `/map fracture` on an existing Fracture dumping a second set of ingress tokens.
- **`Disable Fracture`** button in `/fow fow_options` → OTHER. Reads/writes the existing `noFractureMode`, so it and `weird_game_setup no_fracture` can't diverge.

(FOW) BUG FIXES:
- **Fog ingress actually works** — three bugs: GM-routed buttons carried the player's `FFCC_` prefix so **the GM couldn't press them**; the "wait for the GM" notice fired once per tech type (up to 6×); tech types with few candidate tiles were skipped entirely, so those tokens were never placed by anyone.
- **Planesplitter could conjure an ingress from nothing** — step 1 adds a token, step 2 removes one, and neither trigger checked the Fracture was in play. Guarded in `initializePlanesplitterStep1` (covers strategic action, tech action, Al'Raith, Unravel).
- **Blocked effects now say so** — Silver Flame, Planesplitter gain, Unravel and Cabal BT were silent no-ops, and two then offered ingress buttons that couldn't resolve.
- **Egress is data-driven** — `FoWHelper` used an `egress*` alias prefix, now uses `hasEgress` (alias kept as fallback).

-> Important Review points:

- **`no_fracture` now blocks every entry path, not just rolls.** It said "turn off fracture rolling" but also hid the frac row — which only makes sense if the Fracture is never present. Pre-change an automatic effect could still force it in, and it rendered overlapping the galaxy on an unresized canvas. Roll-only wasn't a coherent contract. Description updated.
- **Rendering decoupled from `no_fracture`** — the frac row draws whenever it's on the board, so that overlap state is now unreachable, including for saves already in it.
- **The 9 `cpti20*` Cosmic Phenomena tiles now count as Fracture** (they carry the fracture card back and are clearly a parallel region). So in a CP game the TE Fracture won't spawn on top, and fracture-keyed effects apply inside them. `cpti20fracxus` also gained `hasEgress` — its `fracxusegress` alias never matched the old check, so that region had no way in at all.
- **`token_fracture_async.png` is placeholder art.** Correct size (95×95 RGBA, matching nebula/asteroids/gravity rift/ingress) but should be replaced. It's load-bearing, not cosmetic — `/add_token` rejects a token whose file doesn't exist.

-> Method: 
- Wrote original implementation. Token and Fog guarding
- used claude to identify weaknesses
- wrote fixes
- had claude improve my edits and sugegst the API for all the call sides and help with merging frac position and new isFraction
- automated testing 3 fault
- manual testing in test bot environment.

-> Tests

- `mvn test` — 558 pass. New: `TileFractureTest` (flag works wherever the tile sits, legacy `frac` positions, **unknown tile IDs don't NPE**, token path, egress, and the full `/add_token` resolution chain so missing art fails the build); `TileModelTest` (every fracture-backed tile must carry the flag).
- Manual review all changes
- Independend NN reviews by Codex, GPT5.6 and Claude

## Manual testing

Done:

- `/add_token token:fracture tile_name:101` — first failed with `Token: fracture is not valid`, which is what surfaced the missing-art requirement
- `/tech add planesplitter-obs` — correctly does *not* spawn; only `planesplitter-firm` / `tf-planesplitter` have the gain clause. This is what led to finding the unguarded ingress trigger

- [x ] `/map fracture` places tiles + units + ingress; frac row renders correctly; second call refused with no extra ingress
- [x ] `/add_token fracture` on a normal system makes it Fracture space and blocks a spawn
- [x ] `no_fracture` and the FoW button are the same setting (`/game info` reflects both)
- [x ] with it on: mandatory effects explain themselves, no dead ingress buttons
- [x ] fog: player gets one wait-for-GM ping; GM room gets one message per tech type
- [x ] fog: **a GM who isn't the triggering player can press the ingress buttons**
- [x ] fog: only players who can see the system get the reveal
- [x ] `planesplitter-obs` with no Fracture: reports "not in play", offers nothing


src/main/java/ti4/service/map/FractureService.java
```diff
@@ -72,8 +91,21 @@ private static void resolveFractureRoll(ButtonInteractionEvent event, Game game,
...
     }
 
-    public static void spawnFracture(GenericInteractionCreateEvent event, Game game) {
-        if (isFractureInPlay(game)) return;
+    /** Brings The Fracture into play for an automatic effect, or says why it could not. */
+    public static boolean enterPlayOrExplain(
+            GenericInteractionCreateEvent event, Game game, @NotNull Player player, String breakthrough) {
+        if (!spawnFracture(event, game)) {
+            String why = whyFractureCannotEnterPlay(game);
+            if (!why.isEmpty()) MessageHelper.sendMessageToChannel(player.getCorrectChannel(), why);
```
Fingon00: In non-Fog games, there should be no message telling us that the fracture is already in play. Such a message will be useless/redundant/annoying. It should just return in that case

Stabar-ti: Updated to be less noisy. It now only throws teh message (in teh 5 automated cases, silver flame cabal hero etc.) when the fracture is disabled to come into play. (Poor cabal and firnament)

PR #5539, opened 2026-08-17
Fog setup helper scheme

Stabar-ti:
# GM setup wizard for Fog of War games (`/fow setup`)

**EDIT: most new lines are explain texts and commented text for me to work the code.**

Also contains 2 new API calls for list handling, as the existing ones were not complete

## Summary

- Adds `/fow setup`, a GM-only 9-step wizard that walks a Fog of War game from creation to first turn in the GM room. Steps: **Game Type → Game Options → Map → Decks → Factions → Table Order → Neutral Player → Review → Done**. State lives in one `FowSetupWizardState` object persisted via `StoredValueHelper`, so the panel survives restarts and the GM can navigate back and forth.
- The panel is a single self-editing message (`Step N/9` header) rather than a growing stack of button posts. Each step also creates a `{game}-setup-{step}` info thread once, GM-role pinged, explaining what that step does and what the non-obvious options mean.
- **Game Options** groups homebrew/faction/deck/scoring settings into four category submenus of clear on/off toggles (`✅`/`🚫`), each with a Done button so submenus collapse instead of piling up. Includes Fractured TIGL and Alliance mode (mutually exclusive).
- **Factions** supports manual assignment, a mini faction deal, GM-driven faction bans, and Alliance pairing. 
- **Table Order** offers manual seat picks, per-player private dice rolls (nobody sees another's roll), ascending/descending resolution, and a Set Speaker button. 
- **Positions are assigned last, by design.** Loading a map (string or JSON) and starting a draft both wipe placements, so the wizard gates placement behind "a map exists" while still allowing drafting and faction assignment beforehand. Unplaced/temp-anchored players (Franken parks them at off-map `50x` anchors) are detected and re-offered.
- A **FoW Setup Wizard** button is now posted in the GM room whenever a FoW game is created, via both `/fow create_fow_game_button` and the `/bothelper` variant.

-> Cleanup is explicit: **Close Wizard** deletes the panel, archives (not deletes) the info threads, and sweeps any open ban-list pages; `/game end` clears the stored state. Nothing is left behind on the server or in save data.

### Twilight's Fall in FoW

Milty and Nucleus drafts are no longer offered in FoW games — they draft slices, tiles and speaker order, all of which the wizard handles itself, and tile/draft-order categories are already forced to 0 in FoW. In their place FoW gets **Inaugural Splice Only**, so players still draft abilities/units/genomes. That path deliberately calls `setUpFrankenFactions(..., force=false)` so it doesn't re-park players the GM has already placed.

**## Touched outside FoW **

Every change below is either additive, a pure extraction, or FoW-gated. No non-FoW behaviour changes.

| File | Change | Why non-FoW is unaffected |
|---|---|---|
| **`MessageHelper`** | +64 lines, **zero removed** — two new shared helpers (`postOrEditWithButtons`, `sendPagedSelectMenus`) | Purely additive. No existing method touched, so nothing can regress; both derive their limits from JDA constants (`SelectMenu.OPTIONS_MAX_AMOUNT`, `Message.MAX_COMPONENT_COUNT`) instead of hardcoded `25` |
| `**FrankenDraft` + 4 subclasses** | New `isFowExcludedCategory` (BLUETILE/REDTILE/DRAFTORDER → 0), applied in each variant's limit lookup | The helper's first condition is `getOwner().isFowMode()`, so outside FoW it returns false and the original limit is returned unchanged |
| **`FrankenDraft`, `FrankenDrazDraft`** | `getDraftableFactionsForGame` widened `private static` → `public static` | Visibility only, no body change. Also fixes a latent inconsistency: the subclass was hiding the parent's method at reduced visibility |
| **`HomebrewService`** | Extracted `applyHomebrew(...)` out of the button handler; three fields `final` → `public final` | The handler now delegates to the extracted method — same code, same order. The split exists so the wizard can reuse the logic without inheriting the handler's delete-the-clicked-button side effect |
| **`WeirdGameSetup`** | Extracted `applyThundersEdgeMode(...)`; TF homebrew posts now route via GM room when `isFowMode()` | Extraction is line-for-line identical. The channel change is behind `game.isFowMode() ? … : game.getMainGameChannel()` — non-FoW keeps the main channel |
| **`TEOptionService`** | New `homebrewChannel(game)` replacing hardcoded `getMainGameChannel()` in 8 spots | Same ternary pattern — identical channel outside FoW. This was a genuine FoW leak: the command has no FoW gating, so it was reachable on any game |
| **`SetDeckService`, `DeckConfigImportService`** | Signatures widened `SlashCommandInteractionEvent` → `GenericInteractionCreateEvent`; `resetDeck` lifted to a parameter | Slash-command callers pass the same values they read before; the import path passes `false`, mirroring its prior always-false behaviour (it never had a `reset_deck` option) |
| **`FrankenSettings`, `AutoCompleteProvider`** | Inaugural Splice hidden as a *standalone* draft mode outside FoW | Behaviour **fix**, not a regression: outside FoW the splice only ever runs as an automatic follow-up to a milty/nucleus draft, so picking it as an opening draft dealt a bag with no factions, tiles or home systems |
| **`EndGameService`** | Clears wizard state on game end | Guarded by `if (game.isFowMode())` |
| **`StatusHelper`** | +4 lines, **comment only** | A `TODO FOG LEAK` note documenting an unconditional planet reveal to the main channel. Zero executable change; flagged for a separate PR rather than fixed here |

## TIGL

Two things, both surfaced by live testing rather than by the wizard's own code.

**Fixed an NPE that crashed TIGL setup.** `Player.getUser()` returns `null` for any id JDA can't resolve — which includes the neutral "Dicecord" dummy that every FoW game carries — and `TIGLHelper` dereferenced it. There were two null paths, not just the one in the stack trace: `allUsersAreMembersOfHubServer` and the rank-snapshot loop. `setTIGLRankSnapshotAtSetup` now filters dummies into a `rankedPlayers` list used by both, plus a defensive null guard. The dummy filter matters beyond the crash: a bot account isn't a league participant, and counting it would have permanently reported the game as containing non-hub members and silently disabled rank handling. Also added an empty-list guard — `getLowestCommonRankBetweenPlayers` seeds at the *top* rank and walks down, so an all-dummy game would have stamped `Archon`/`Hero` as the game's minimum rank and drawn that badge on the map.

**Blast radius checked**: non-FoW TIGL games have no dummies, so `rankedPlayers` is exactly `getPlayers().values()` and behaviour is identical. Eliminated players do become dummies (`ButtonHelper`), but every caller of `initializeTIGLGame` is creation/setup-time, so elimination can never retroactively alter a snapshot.

**In the wizard, only Fractured TIGL is offered.** Standard TIGL is never legal in FoW, so there is deliberately no plain-TIGL button. Enabling Fractured TIGL turns Alliance mode off and vice versa.

**NOT FIXED, reported for a follow-up:** Fractured TIGL is stored as a *tag string* (`"TIGL Fractured"`), so `/game tags` can write it directly with no validation. That yields a half-configured game — the tag flips but `competitiveTIGLGame`, the rank snapshot and the setup text don't. Worse in reverse: dropping the tag from a game that *is* competitive TIGL silently demotes it to standard, and the incompatible-mode guard lives in `setCompetitiveTIGLGame`, which a tag edit never calls — so a FoW game can end up flagged as standard competitive TIGL. `TiglReportService.determineLeague` reads the tag alone, with no `isCompetitiveTIGLGame` check. The wizard itself is robust (it requires `isFracturedTIGLGame(game) && game.isCompetitiveTIGLGame()`), so this is pre-existing and out of scope here.

## Test plan

- [x] `mvn test` — **585 tests, 0 failures, 0 errors** (up from 582; +3 new `TIGLHelperTest`)
- [x] `mvn spotless:apply` clean
- [x] New `TIGLHelperTest`: dummy players don't crash or disable rank handling; an *unresolvable real player* still disables it (guard rail against widening the dummy filter); an all-dummy game records no rank. **Verified these fail without the fix** — reverting it reproduces the exact production error (`Cannot invoke "User.getId()" because "user" is null`); reverting only the empty guard yields `expected: <null> but was: <archon>`
- [x] New `FowSetupFactionServiceTest` (GM gating on modal handlers) and `TEOptionServiceTest` (FoW hides Milty/Nucleus, offers splice)
- [x] `git merge-tree master HEAD` reports clean — no conflict with master's newer `TIGLHelper` work (#5536)
- [x] Live pass against `.idea/FOW_SETUP_WIZARD_TESTING.md` Tier 1: every button acknowledges (no "This interaction failed"), edit-in-place doesn't stack duplicates, gate-faction double setup (Creuss + Crimson), TF position assignment, Custom Scoring modal opens
- [x] Live FoW containment sweep: read `{game}-anonymous-announcements-private` top to bottom after a full wizard run.
- [x] Live teardown: Close Wizard archives threads and sweeps ban pages; `/game end` deletes the category and clears stored state

## Known gaps (not bugs)

- Positions can't be **re-assigned** once a player is truly placed; only unplaced/temp-anchored players are offered.
- Red Tape / Ignis Aurora / Omega Phase aren't in the Game Options categories — reach them via Step 1 → **Supported Homebrew**.
- In a Twilight's Fall game, Crimson/Ghost set up through the wizard **don't receive** `echo_of_divergence` / `echo_of_sacrifice`. These are granted only in `MiltyService.setupExtraFactionTiles`, which `PlayerSetupService` doesn't call — pre-existing upstream, inherited here. <--- did not had any idea if this is a real bug!!!?
- The gate→home mapping has no single source of truth (encoded in ~6 logic sites and ~4 rendering sites, in two divergent implementations). The real fix is a `FactionModel` field for the second home system; deliberately out of scope.
- `Dev: Test Tiles` is a developer shortcut visible to all GMs.


src/main/java/ti4/helpers/TIGLHelper.java
Rabats, line 1: The most widespread damage potential

Rabats: ignore

Rabats (approved):
If something breaks something wider it will be TIGL but i triple checked

PR #5541, opened 2026-08-17
Beginning of Deep Reaches Rework

IslandForShans:
Reworked Dreaming Throne

- Dream Nexus now makes the tile a nebula for all players, dream player movement is unaffected
- Visions promissory gives nebula movement negation to holder
- Changed dream commander unlock
- Blue tech now both gives the +1 to movement starting in nexus token system and makes nebulae adjacent for them.
- BT changed to a Blue<>Green Synergy

Most of the code is moving it into separate handler files to match the rest of my factions. These will appear to have a lot of code changes, trying to keep it minimal and succinct for each one.


IslandForShans:
Forgot I initially changed all the abilities in the ability JSON, so this may end up being a bigger PR than I wanted....

I'll see about fixing it tonight.

PR #5555, opened 2026-08-21
FOW: Planet select service leak fixes and migration

Stabar-ti:
# Fog of War: planet targeting leaked target's holdings

**Bug/Leak:** in FoW games, planet-targeting cards (Cripple Defenses, Elect Planet agendas, etc.) built their
button lists by looping over `target.getPlanets()` — safe outside fog, a holdings leak inside it. Same shape
in 20+ places.

**Fix:** one entry point, `PlanetTargetService`. A planet is offered only if the actor could know it exists
(sees it now, has ever seen it, or can see the owner's stats). Everything else goes through Blind Target
(type a name); an illegal or off-map target fizzles silently at resolution — never a reason.

## Scope
- 21 components migrated. New `Ownership` / `Visibility` axes replace two unrelated booleans.
- Also fixed, found during the sweep:
  - `AgendaRiderHelper` rider button missing its `FFCC_` ownership gate — any player could press another
    player's rider target. Affects fog **and** non-fog Franken-draft games (shared code path).
  - Two stale-button crashes (`VoteButtonHandler` on a closed agenda window, `echoPlaceFrontier` on an
    off-map Blind Target position) -? both now fizzle instead of throwing. Existence-guard-only; can't change
    behavior for a well-formed button.
- two **rules bypasses**, not just leaks:
  - **Settlements**: the fog list's comment claimed the resolution handler checked "planet controlled by a
    voter for the winning outcome" . A fog player could coexist on *any* known non-home planet,
    not just a winning voter's. Now enforced at resolution.
  - **Kairn agent**: same shape -> fog offers all three explore-trait decks (correct, avoids disclosing the
    target's planet traits) but the claimed downstream eligibility check didn't exist. A fog player could
    place a card from a trait deck the target doesn't hold. Now enforced at resolution.
  - **Ragh's Call**: `Planet.isSpaceStation()` and `isSpaceStation(Game)` disagree on one case (a Gledge Core
    planet in Twilight DS). My resolution guard used the game-arg overload unconditionally while the non-fog
    builder used the no-arg one, so a Twilight DS non-fog game would silently fizzle a destination that used
    to work. Now mode-matched to whichever overload built the button.
  - Common thread on all three: a comment asserting "checked downstream" where no such check existed. Worth
    distrusting that exact phrasing on any future pass.
- **Out of scope:** 6 agendas that mis-route on substring matching — needs new outcome flows, separate PR.

## Space stations in fog

No single rule — it's a per-card fact, same as pre-refactor. Confirmed per component:

| Component | Fog rule | Matches non-fog |
|---|---|---|
| Cripple / Uprising / Plague / Reparations / Unstable | not filtered | yes |
| Reactor Meltdown | not filtered (only off-map pseudo-planets + home excluded) | yes |
| Elect Planet agenda vote | excluded | yes |
| Agenda rider planet list | not filtered | yes (pre-existing gap, unchanged) |
| Ragh's Call | excluded | yes, now mode-matched (see above) |
| Yin hero | excluded only in the owned branch | yes, by design (the original round-1 regression) |
| Stellar Atomics | excluded via the game-arg overload | fog-gated only — a Gledge Core planet in Twilight DS is fog-excluded but non-fog-included; not a bug (can't touch non-fog), just an inconsistency on an obscure combo, left as-is |

## Testing
- Unit: 610 tests, 0 SpotBugs (`mvn clean verify -P ci-spotless`). Includes a structural leak-guard scan over
  the rest of the codebase and a per-card spec-pin test so a future refactor can't silently drop a rule.
- Live, via `/developer test_fog_planet_targets` — restored twice to cover new findings, now removed again
  and archived at `.idea/archived-dev-tests/TestFogPlanetTargets.java` (not in the diff). Its fourth panel,
  **Rules-bypass regressions**
  - Also manually tested 3 key ones in a full live escalation tree.


src/main/java/ti4/discord/interactions/buttons/handlers/agenda/VoteButtonHandler.java
```diff
@@ -90,20 +110,33 @@ private static void firstStepOfVoting(Game game, ButtonInteractionEvent event, P
         }
     }
 
+    // Non-fog only: the fog path never offers the player-picking step that leads here.
```
Fingon00: THen why are you adding a check for fog here?

src/main/java/ti4/discord/interactions/buttons/handlers/agenda/VoteButtonHandler.java
```diff
@@ -90,20 +110,33 @@ private static void firstStepOfVoting(Game game, ButtonInteractionEvent event, P
...
     @ButtonHandler("planetOutcomes_")
-    static void planetOutcomes(ButtonInteractionEvent event, String buttonID, Game game) {
+    static void planetOutcomes(ButtonInteractionEvent event, String buttonID, Game game, Player player) {
         String factionOrColor = buttonID.substring(buttonID.indexOf('_') + 1);
         Player planetOwner = game.getPlayerFromColorOrFaction(factionOrColor);
         String voteMessage = "Choosing to vote for one of " + factionOrColor
                 + "'s planets. Please use the buttons to choose the planet you wish to vote for.";
         List<Button> outcomeActionRow;
-        outcomeActionRow = getPlanetOutcomeButtons(planetOwner, game, "outcome", null);
+        if (game.isFowMode()) {
```
Fingon00: If fog never comes here, what is the purpose of all this code?

src/main/java/ti4/discord/interactions/buttons/handlers/agenda/VoteButtonHandler.java
```diff
@@ -113,14 +146,62 @@ static void tiedPlanets(ButtonInteractionEvent event, String buttonID, Game game
                 + "'s planets. As Speaker, please decide a winner.";
 
         List<Button> outcomeActionRow;
-        outcomeActionRow = getPlanetOutcomeButtons(planetOwner, game, "resolveAgendaVote_outcomeTie*", null);
+        if (game.isFowMode()) {
+            // Reached when nobody voted, and from manual resolution. "Any planet" is rules-correct here, but
```
Fingon00, line 146: This comment is not accurate. Tied planets is also reached when several planets were voted on and are tied. Those planets could be planets the speaker does not know of. This code seems broken.

src/main/java/ti4/discord/interactions/buttons/handlers/faction/base/yin/YinHeroButtonHandler.java
```diff
@@ -37,6 +56,7 @@ public static void yinHeroStart(ButtonInteractionEvent event, Game game) {
                 event.getChannel(), "Please choose the player that owns the planet you wish to land on.", buttons);
     }
 
+    // Non-fog only - see yinHeroStart, which skips this step entirely in fog.
```
Fingon00, line 59: Then why the fog code later in the method?

src/main/java/ti4/discord/interactions/buttons/handlers/faction/homebrew/whispers/vyserix/VyserixLeaderHandler.java
```diff
@@ -12,17 +12,34 @@
...
+import ti4.service.fow.PlanetTargetService.PlanetTargetSpec;
 import ti4.service.unit.AddUnitService;
 
 @UtilityClass
 public class VyserixLeaderHandler {
 
     public static void offerHeroAttachmentButtons(GenericInteractionCreateEvent event, Game game, Player player) {
         List<Button> buttons = new ArrayList<>();
+        String message = player.getRepresentation()
+                + ", choose a non-home planet without a technology specialty to attach _Titles Are Silly_ to.";
```
Fingon00, line 28: I know this wasnt your work originally, but are you not reading what this is saying?

src/main/java/ti4/discord/interactions/buttons/handlers/agenda/VoteButtonHandler.java
Stabar-ti: It was there because as a fallback. Never should be: never should reach there. And I was 99% confident it will not get to that point... However, I will remove it as a second look at it would only happen in a case that does not "exist yet unnecessary guarding from my side.

src/main/java/ti4/discord/interactions/buttons/handlers/agenda/VoteButtonHandler.java
Stabar-ti: see comment above

src/main/java/ti4/discord/interactions/buttons/handlers/agenda/VoteButtonHandler.java
Stabar-ti: Speaker will get offered the planets that were voted on to break the tie. -> The planets don't need to exist or he needs to know about. When someone votes for a planet that planet will be auto added to your list of known planets. I'm pretty sure the code here is correct.

src/main/java/ti4/discord/interactions/buttons/handlers/faction/base/yin/YinHeroButtonHandler.java
Stabar-ti: I never touched YinHeroTarget_

The Fog code is in the next method. -> yinHero-Planet. The diff does not show the break between the two methods.

src/main/java/ti4/discord/interactions/buttons/handlers/faction/homebrew/whispers/vyserix/VyserixLeaderHandler.java
Stabar-ti: I read it, yes. :D Thought it was a funny easter egg

Stabar-ti:
Fixed 2 holes pointed out by the review. The Fix was more extensive for the 25 button limit than i anticipated. Also fixed the blind selection button to follow rules.....

Stabar-ti:
needed a conflict solve

PR #5577, opened 2026-08-27
LL bug fixes

IslandForShans:
Added supermassive purge handling to normal fragment purge paths (only runs when the id has "supermassive" in the id.

Changed aeterna graveyard I and II to let the player enforce the produced unit.


IslandForShans:
Oh also added a `/relic show_purged_fragments` command

PR #5581, opened 2026-08-28
Deep reaches Ta rework

IslandForShans:
Full rework of the Ta


IslandForShans:
Well this is a wee bit frustrating. I'll need to take a closer look later.

PR #5684, opened 2026-09-20
Give Theodisi's Exploration Rider its own alias

wholton:
Theodisi and Action Deck 2 both defined an action card aliased `exploration_rider`, with different effects:

| Source | Text |
| --- | --- |
| `theodisi` | You cannot vote on this agenda. Predict aloud the outcome of this agenda. If your prediction is correct, **explore 1 planet you control of each trait**. |
| `action_deck_2` | You cannot vote on this agenda. Predict aloud an outcome of this agenda. If your prediction is correct, **explore up to 3 different non-cultural planets you control**. |

`Mapper` loads every file under `data/action_cards` into a single map keyed by alias, so only one of the two definitions survived startup and *both* decks served whichever one won. Four Theodisi decks (`action_cards_pok_te_ll`, `..._us_ll`, `..._us_br_ll`, `..._br_ll`) and six ACD2-based decks all pointed at the same entry.

### Change

The Theodisi card becomes `exploration_riderll`, following the existing convention for Theodisi's duplicated cards (`fsll` → Flank Speed, `dhll` → Direct Hit, `saboll` → Sabotage), and gains `"automationID": "exploration_rider"` so it still routes to the same automation — the same pattern those cards already use. The four Theodisi deck lists are updated to match; the ACD2 decks are untouched and keep `exploration_rider`.

### Why the automation still works

`ActionCardHelper` already distinguishes the two variants when naming the rider:

```java
String riderName = actionCard.getSource() == ComponentSource.theodisi
                && "exploration_rider".equals(actionCard.getAutomationID())
        ? "Exploration Rider (LL)"
        : actionCardTitle;
```

That check reads `source` and `automationID`, not the alias, so it is unaffected — the Theodisi card still matches both conditions, and the ACD2 card still fails the source check. Downstream resolution in `AgendaHelper` keys off the `"Exploration Rider (LL)"` display name, and `ExplorationRiderLLButtonHandler` uses its own button id rather than the alias, so neither is touched.

### Verification

All action card and deck JSON parses; alias duplicates across `data/action_cards` drop from 1 to 0; every card ID referenced by the 30 action-card decks resolves. Theodisi decks now resolve `exploration_riderll` and ACD2 decks `exploration_rider`.

No overlap with #5683 — that PR touches `acd2.json`, `legacy_data.json`, `action_deck_2.json` and `other_decks.json`; this one touches only the two `theodisi.json` files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


wholton:
Folded into #5683 — the same commit now rides along there, so this is redundant. Closing.

